PR #16, opened 2020-03-06
adding in intake-geopandas

rsignell-usgs:
adding in intake-geopandas , required to run agu2019 tutorial


scottyhq:
added to pangeodev/pangeo-notebook:2020.03.11

PR #19, opened 2020-03-27
added a couple libraries to pangeo-notebook

scottyhq:
testing PR workflow


scottyhq:
Should create a separate PR workflow per notebook to only run if files have changed

PR #21, opened 2020-03-29
Minor improvements

tjcrone:
This PR:
 - adds .yml to the base-image/condarc filename so that it gets proper syntax highlighting
 - copies configuration files in a new layer for clarity
 - silences the apt-utils warning (see: https://github.com/phusion/baseimage-docker/issues/319)
 - uses the default name for the Conda lock file to make it consistent with upstream examples
 - updates and fixes a few typos in README.md

Hopefully this is helpful! More ideas coming. Thanks @scottyhq.


scottyhq:
Thanks @tjcrone ! This all looks good. I originally went with the 'spec-file.txt' name to match conda documentation, but I also like the more explicit `conda-linux-64.lock` name - see https://github.com/mariusvniekerk/conda-lock/issues/10. I'll go ahead and merge and retag with todays date.

base-image/Dockerfile
```diff
@@ -62,6 +67,7 @@ ONBUILD USER root
 # hardcode for now
 ONBUILD COPY --chown=jovyan:jovyan . /home/jovyan
```
herculan0, line 68: Can you change this for widespread use?

PR #24, opened 2020-03-30
use miniforge, makefile updates

### LINKED ISSUE #14: use miniforge to install conda base environment
Following @yuvipanda 's lead in https://github.com/jupyter/repo2docker/pull/859 
> miniforge is a new community-led installer that uses conda-forge as the default
channel, rather than defaults. So no mixing of defaults & conda-forge channel by default. and would likely reduce image size.

@TomAugspurger also looked into using just the conda standalone executable since we don't really need a (base) environment, but it is what people are used to locally, so I think it is worth keeping.
https://github.com/TomAugspurger/pangeo2docker/blob/f687f97dcf8f48d34033ea1903befd12d578f659/pangeo2docker/Dockerfile.tpl#L25

### LINKED ISSUE #15: Add option to use conda packages.txt or conda.lock file in addition to environment.yml
The current setup only supports conda environment.yml files (https://github.com/pangeo-data/pangeo-stacks-dev/blob/master/pangeo-notebook/environment.yml), consistent with repo2docker (https://repo2docker.readthedocs.io/en/latest/config_files.html#environment-yml-install-a-python-environment)

It would be nice to also support alternative conda files for more flexible `conda create` commands instead of `conda env create -f environment.yml` For example it would be great to utilize 

1) new `conda.lock` files pointed out by @ocefpaf 
https://github.com/mariusvniekerk/conda-lock#dockerfile-example 

2) conda `packages.txt` from `conda list --export > packages.txt`
```
# This file may be used to create an environment using:
# $ conda create --name <env> --file <this file>
# platform: osx-64
anaconda-client=1.7.2=py37_0
asn1crypto=1.3.0=py37_0
```

these files require an extra step to generate but lead to more reproducible environments.

### LINKED ISSUE #22: Review of Dockerfile cleanup commands
Currently we are running a lot of custom cleanup commands to save space on the images:
https://github.com/pangeo-data/pangeo-stacks-dev/blob/899e2e7bcc76354f1d4b584dc21742492cd81325/base-image/Dockerfile#L92-L96

I just ran into an error where one of the commands failed, possibly related to a change in layout with bokeh 2.0. cc @TomAugspurger . I'll go ahead and comment it out for now
```
bokeh-2.0.0          | 6.8 MB    | ########## | 100%
find: ‘/srv/conda/envs/pangeo/lib/python*/site-packages/bokeh/server/static’: No such file or directory
```

scottyhq:
cc @tjcrone @TomAugspurger 

* This moves us to miniforge instead of miniconda, closes #14
* Since we've done a bit of testing now with either `environment.yml` or `conda-linux-64.lock` I'll also close #15 
* Also adds a fix to close #22 (always install user packages in "notebook" environment


TomAugspurger:
Thanks. No thoughts on miniforge vs. miniconda. I don't know how stable / mature miniforge is.

scottyhq:
> Thanks. No thoughts on miniforge vs. miniconda. I don't know how stable / mature miniforge is.

Me neither really, just following repo2docker/mybinder.org's lead: 
https://github.com/jupyter/repo2docker/pull/859

PR #33, opened 2020-03-31
try chatops for condalock

scottyhq:
This addresses some pieces of https://github.com/pangeo-data/pangeo-stacks-dev/issues/27. In particular it will enable generating lockfiles from PRs using "ChatOps" (https://github.com/peter-evans/slash-command-dispatch). 

So in theory someone edits environment.yml in there fork and creates a PR. Then in a PR comment writes:

/condalock

Then a bot will run conda-lock and add the lockfiles for each image into the PR. I'll probably follow up with some direct commits to test this out or additional PRs.

### DIFF
--- a/.github/workflows/ChatOps.yml
+++ b/.github/workflows/ChatOps.yml
@@ -1,0 +1,17 @@
+# https://github.com/peter-evans/slash-command-dispatch
+name: Slash Command Dispatch
+on:
+  issue_comment:
+    types: [created]
+jobs:
+  slashCommandDispatch:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Slash Command Dispatch
+        uses: peter-evans/slash-command-dispatch@v1
+        with:
+          token: ${{ secrets.PERSONAL_TOKEN }}
+          reaction-token: ${{ secrets.GITHUB_TOKEN }}
+          issue-type: pull-request
+          permission: none
+          commands: condalock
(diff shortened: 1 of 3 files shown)


tjcrone:
Why do we not just want to build conda-lock on a PR from a fork?

scottyhq:
@tjcrone I think this is the only way to add a commit to someone's PR.

tjcrone:
Can we add a commit on a merge?

scottyhq:
Probably, but the problem with that is that you can't really run accurate tests before merging (since building uses the lockfile). 

I can imagine scenarios where the lockfile generation fails or the image build fails with the lockfile for some reason. Then we'd have to follow up with additions commits. 

It would be nice to keep the commits to master to a minimum and guarantee that for any commit to master `environment.yml -- conda-linux-64.lock -- and dockerimage:comit` are always in sync.

tjcrone:
Can we build the lock file for the tests but not commit it until the merge?

tjcrone:
The reason I am asking here is that while chatops are pretty cool, the goal here was to simplify, and I don't see chatops, which almost nobody has even heard of much less used, as a simplification. I think keeping the commit history clean with an extra commit on merge seems a lot better and simpler to me than chatops. Maybe someday everyone on GitHub will be chatoping left and right. But at this stage, it seems pretty unconventional and likely to add confusion.

PR #41, opened 2020-04-01
automatically run condalock workflow for watchcondaforge action

scottyhq:
new metapackages on conda-forge will now trigger a PR - https://github.com/pangeo-data/pangeo-stacks-dev/pull/40 

this small change should trigger the separate condalock workflow, which solves for the conda environment, commits package lock files, then rebuilds images with those lock files and runs tests.

cc @jhamman

### DIFF
--- a/.github/workflows/WatchCondaForge.yml
+++ b/.github/workflows/WatchCondaForge.yml
@@ -37,6 +37,7 @@ jobs:
           title: "Update pangeo-notebook metapackage version to ${{ steps.latest_version.outputs.version }}"
           reviewers: "scottyhq"
           branch: "upgrade-pangeo-metapackage-version"
+          branch-suffix: timestamp
           body: |
             /condalock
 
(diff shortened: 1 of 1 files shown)


.github/workflows/WatchCondaForge.yml
```diff
@@ -37,7 +37,12 @@ jobs:
           title: "Update pangeo-notebook metapackage version to ${{ steps.latest_version.outputs.version }}"
           reviewers: "scottyhq"
           branch: "upgrade-pangeo-metapackage-version"
+          branch-suffix: timestamp
```
jhamman, line 40: I'm not sure we want this. Couldn't this lead to an accumulation of PRs (say over a weekend)?

scottyhq: i don't think we'll be updating the metapackages that regularly. But worst case scenario we just close a couple of PRs. Or merge each separately so that there is a corresponding image for each metapackage version. Otherwise commits get force-pushed to the same PR and the image building wouldn't automatically retrigger because a new comment isn't made (see https://github.com/scottyhq/githubactions-test/pull/1)

PR #46, opened 2020-04-02
avoid force pushing to existing repo

scottyhq:
The current WatchCondaForge.yml force-pushes to an existing PR every hour, retriggering CI and sending emails. https://github.com/pangeo-data/pangeo-stacks-dev/pull/44 . This should avoid force pushing if an automated 'upgrade-pangeo-metapackage-version' PR already exists. cc @jhamman

### DIFF
--- a/.github/workflows/WatchCondaForge.yml
+++ b/.github/workflows/WatchCondaForge.yml
@@ -60,6 +60,7 @@ jobs:
             Updated `environment.yml`s to use `${{ steps.latest_version.outputs.version }}`.
 
       - name: Add Condalock Comment
+        if: steps.check_branch.outputs.exists == 'False'
         uses: peter-evans/create-or-update-comment@v1
         with:
           token: ${{ secrets.PERSONAL_TOKEN }}
--- a/.github/workflows/WatchCondaForge.yml
+++ b/.github/workflows/WatchCondaForge.yml
@@ -5,6 +5,9 @@ on:
   schedule:
     - cron: "0 * * * *"
 
+env:
+  PR_BRANCH: upgrade-pangeo-metapackage-version
+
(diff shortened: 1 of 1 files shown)


.github/workflows/WatchCondaForge.yml
```diff
@@ -29,21 +32,35 @@ jobs:
...
+          git fetch --no-tags --prune --depth=1 origin +refs/heads/*:refs/remotes/origin/*
+          if git show-ref --quiet refs/remotes/origin/${PR_BRANCH}; then
+            RESULT=True
+          else
+            RESULT=False
+          fi
+          echo "::set-output name=exists::${RESULT}"
+
       - name: Create Pull Request
+        if: steps.check_branch.outputs.exists == 'False'
```
jhamman, line 48: Do we need this line? I understand desire to put this condition on the `condalock` command but it would be nice to keep updating the PR as new packages become available.

scottyhq: thinking about this a bit more I think this workflow is better suited to https://github.com/peter-evans/repository-dispatch rather than a cronjob. Since we have control of the metapackage repository we can trigger something here when a new version is released instead of polling every hour? I'll try here https://github.com/scottyhq/githubactions-test and then revisit this PR

scottyhq:
closing this, can revisit later if necessary

.github/workflows/WatchCondaForge.yml
scottyhq: unfortunately https://github.com/peter-evans/repository-dispatch requires a personal access token, and since we can't add secrets to https://github.com/conda-forge/pangeo-notebook-feedstock that different approach won't work

PR #49, opened 2020-04-07
force rebuild to get latest panel 

rsignell-usgs:
/condalock

Really just kicking a new build to get latest "panel" 0.9.5 build 1 on conda-forge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ conda env create -f environment-condalock.yml
 git checkout -b change-pangeo-notebook
 ```
 
-edit `pangeo-notebook/environment.yml` to change packages! Note that `make pangeo-notebook` is a convenient shortcut to build and test. See the Makefile for specific commands that are run. For example, you can just run conda-lock and don't have to build and test locally.
+Edit `pangeo-notebook/environment.yml` to change packages! Note that `make pangeo-notebook` is a convenient shortcut to build and test. See the Makefile for specific commands that are run. For example, you can just run conda-lock and don't have to build and test locally.
 ```
 make pangeo-notebook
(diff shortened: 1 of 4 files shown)


ocefpaf:
@scottyhq can we do that slash-command-dispatch from an issue?

scottyhq:
Right now we are only setup to run in PRs https://github.com/pangeo-data/pangeo-stacks-dev/blob/e5c550f971ec4064831f7afd8120e13801283186/.github/workflows/ChatOps.yml#L15 .

But that could be a neat ability. Probably would enable a separate command only for administrators such as `/rebuild-all` that could be run from issues. The CI is definitely still a work in progress

scottyhq:
Looks like the build didn't work:
```
CondaHTTPError: HTTP 404 NOT FOUND for url <https://conda.anaconda.org/conda-forge/linux-64/glib-2.58.3-py37he00f558_1004.tar.bz2>
1647
Elapsed: 00:00.195006
1648
CF-RAY: 58053a105d85eaa2-IAD
1649

1650
An HTTP error occurred when trying to retrieve this URL.
1651
HTTP errors are often intermittent, and a simple retry will get you on your way.
```

Despite the message about intermittent failures I ran into this same thing yesterday, so might be an issue with conda / conda-lock

ocefpaf:
That file was moved to broken: https://anaconda.org/conda-forge/glib/files

We need to find a better way to report that. However, the new solve should not find that file :-/

scottyhq:
/condalock

It just built for me locally, so re-triggering build.

rabernat:
I am fascinated by what is happening here and want to understand it better.

What does `/condalock` do?

PR #89, opened 2020-07-01
Add a PANGEO_SCRATCH variable

TomAugspurger:
xref https://github.com/pangeo-data/pangeo-cloud-federation/issues/610,
https://github.com/pangeo-data/pangeo/pull/780. This adds a
PANGEO_SCRATCH environment variable. It relies on the existence of

1. PANGEO_SCRATCH_PROTOCOL
2. JUPYTERHUB_USERNAME

And combines those to form something like
`PANGEO_SCRATCH=gcs://pangeo-scratch/tomaugspurger`

We'll need to update the helm charts to set `PANGEO_SCRATCH_PROTOCOL` to be `s3` or `gcs`. This is a bit convoluted but I couldn't think of a simpler way.


rabernat:
Tom thanks for working on this. I think it is the best solution we can manage right now. We can continue to iterate on this feature if it proves popular.

README.md
```diff
@@ -69,3 +69,13 @@ docker run -it --rm -p 8888:8888 pangeo/base-notebook:latest jupyter lab --ip 0.
...
+
+### Environment
+
+The runtime environment sets two variables by default
+
+1. `$PANGEO_ENV`: name of the conda environemnt.
+2. `$PANGEO_SCRATCH`: a URL like `gcs://pangeo-scratch/username/` that
+   points to a cloud storage bucket for temporary storage. This is set
+   if the variable `$PANGEO_SCRATCH_PROTOCOL` and `JUPYTERHUB_USERNAME`
+   are detected.
```
rabernat, line 81: This seems inconsistent with the start script. The script only checks for PANGEO_SCRATCH_PROTOCOL. If JUPYTERHUB_USERNAME is empty, PANGEO_SCRATCH will get set to `${PANGEO_SCRATCH_PROTOCOL}://pangeo-scratch/`. That could be different from what we want.

base-image/start
```diff
@@ -3,6 +3,11 @@
 # ==== ONLY EDIT WITHIN THIS BLOCK =====
 
 export PANGEO_ENV="base"
+if [[ -z "${PANGEO_SCRATCH_PROTOCOL}" ]]; then
+    :
+else
+    export PANGEO_SCRATCH="${PANGEO_SCRATCH_PROTOCOL}://pangeo-scratch/${JUPYTERHUB_USERNAME}/"
```
rabernat, line 9: I'm thinking it might be better to absorb the bucket name into the variable provided by the hub, rather than hard coded. So something like

```
PANGEO_SCRATCH="${PANGEO_SCRATCH_PREFIX}/${JUPYTERHUB_USERNAME}/"
```

TomAugspurger:
Thanks, updated with your suggestions.

TomAugspurger:
@scottyhq any thoughts here? Otherwise I'll merge in a bit so I can try this out on the new clusters.

Issue #28, opened 2020-03-30
Package Discussion


TomAugspurger:
What packages belong in a "default" pangeo metapackage? Currently `pangeo-notebook` has essentially dask + jupyterhub + jupyterlab. https://github.com/conda-forge/pangeo-notebook-feedstock/blob/master/recipe/meta.yaml. IMO, there's value in having a minimal metapackage.

There's also value in a "useful" metapackage that includes things like

* xarray
* s3fs / gcsfs / fsspec
* intake
* zarr
* matplotlib?
* other viz libraries?

In `pangeo-stacks` we called this `pangeo-notebook`: https://github.com/pangeo-data/pangeo-stacks/blob/a8cf6aefa36800301977390a785d06edac9b915e/pangeo-notebook/binder/environment.yml.

Also, what should we call this? Perhaps just `pangeo`?

cc @rabernat @scottyhq @jhamman

TomAugspurger:
Here's my (deliberately too small to provoke discussion) proposed list

```
  # dask, jupyterlab
  - pangeo-notebook
  # core scipy packages
  - numpy
  - scipy
  - matplotlib-base
  - pandas
  - xarray
  - sparse
  - sympy
  # intake-related
  - intake
  - intake-xarray
  - intake-esm
  - fsspec
  - intake-stac
  # zarr-related
  - zarr
  - gcsfs
  - s3fs
```

Notably absent are

- ML libraries like scikit-learn
- All pyviz libraries
- All geospatial libraries (geopandas, netcdf, cartopy, etc.)
- All xarray-adjacent libraries since I don't know which are generally useful to this community

I think some of those should be included in this kitchen sink package, but I'm not sure which.

jkingslake:
Hi @TomAugspurger 
A few questions:
When you say Xarray-adjacent, do you mean the optional dependencies listed [here](http://xarray.pydata.org/en/stable/installing.html)?

What are the main advantages of keeping the list small? Is that it makes opening the binder faster? Or less likely to run into conflicts?

Personally [nc-time-axis](https://github.com/SciTools/nc-time-axis) would to useful, but I'm not sure how widely its needed. 

From my perspective as someone who is very new to pangeo (and python actually) and who is spreading the word to colleagues with no prior knowledge, there is a big advantage in being able to send someone a link and the binder start up without the need to installing any extra packages. This is an obvious point I guess, but I thought I would emphasize it from my novice point of view.

willirath:
[...]

PR #91, opened 2020-07-02
Add gsw

rabernat:
The oceanographers will need this.

Now that `pangeo-notebook` is the image for our GCP cluster, this is probably the first of many requests we will get for domain-specific packages. Are we okay with that here?

### DIFF
--- a/pangeo-notebook/environment.yml
+++ b/pangeo-notebook/environment.yml
@@ -16,6 +16,7 @@ dependencies:
  - gcsfs
  - geopandas
  - geoviews-core
+ - gsw
  - h5netcdf
  - h5py
  - holoviews


scottyhq:
> Now that pangeo-notebook is the image for our GCP cluster, this is probably the first of many requests we will get for domain-specific packages. Are we okay with that here? 

Good question. I think we need to figure this out ASAP. A key follow-up question is, will oceanographers also be wanting to use the same image on AWS or Azure? Then building here and pushing to DockerHub makes sense to me. That way any hub can run any image built in this repository by using a profile_list (https://github.com/pangeo-data/pangeo-cloud-federation/blob/f2335fbeb84bfb65dcba7190f46dfa4628ac3800/deployments/icesat2/config/common.yaml#L15).

But, I think we should probably do a little refactoring with the image names as discussed in this issue https://github.com/pangeo-data/pangeo-docker-images/issues/28.  Specifically, I think we might want to:

 1. rename `pangeo-notebook` to `oceanography-notebook` and tailor the packages accordingly 
 1. probably create a separate `remote-sensing-notebook` 
 1. change github actions workflows to use those new names

Alternatively, you can switch back to building the `ocean` image in pangeo-cloud federation with hubploy, or `pangeo-notebook` becomes an ever-expanding list of packages used by oceanographers, seismologists, mineralogists... etc ;)

TomAugspurger:
IMO:

1. We build all images here rather than in the hubs. The tooling here is just too nice not to use :)
2. Splitting off to a `pangeo-ocean` image seems fine. The extra overhead of an additional image isn't too big, right?

rabernat:
Splitting up images and building theme here sounds fine to me... 

But if literally the only difference between pangeo-notebook and pangeo-ocean is a tiny package called gsw, it doesn't seem worth the trouble. The GCP hub is used by a wide range of scientists, mostly ocean, atmosphere, and climate folks. So despite what I said in my earlier comment, "oceanography" is probably a bit too narrow.

@scottyhq - I'm curious what additional packages are needed for remote sensing?

I still have a dream that perhaps we don't need a bunch of different images for these very similar domains.

TomAugspurger:
Right I don’t think we’re at the splitting point yet.

scottyhq:
> @scottyhq - I'm curious what additional packages are needed for remote sensing?

@rabernat see for example the environment for the recent icesat2 hackweek https://github.com/ICESAT-2HackWeek/jupyter-image-2020/blob/master/environment.yml

> I push on dask/distributed#3216, to reduce the need for this a bit. But if many people are installing something at runtime then it really should end up in the image. 

This would definitely help. Ultimately I still subscribe to the one streamlined conda environment per project, so am wary of an ever expanding list. It's straightforward to do this for a single notebook server https://gist.github.com/scottyhq/2cbd183374d016940f224e3cb36efe59. Using dask gateway complicates things. I experimented a little bit with a 'bring-your-own-conda-environment' using a dask preload script, but is seems the preference is to have everything baked into a docker image. 


@TomAugspurger - this PR didn't actually make any changes b/c \condalock wasn't run. We need to modify the github action pr.yml to make sure that is run every time.

PR #104, opened 2020-07-26
add geojson extension, rechunker from pip

scottyhq:
/condalock


scottyhq:
/condalock

first command didn't work added as an edit...

PR #105, opened 2020-07-27
Adding cfgrib to notebook env

rsignell-usgs:
/condalock
Adding cfgrib allows us to run the [xarray GRIB example notebook](http://xarray.pydata.org/en/stable/examples/ERA5-GRIB-example.html)

PR #114, opened 2020-08-28
add cfgrib to pangeo-notebook

scottyhq:
replaces #105, trying out modified CI workflow


scottyhq:
i didn't really understand what on `pull_request_target` was doing in the github action. seems it will always get the most recent master commit in this repo, rather than the PR commits. i'll try to restructure once more so that pangeo-bot automatically comments with the /condalock command

PR #124, opened 2020-09-01
add keras to ml-notebook

rabernat:
We need keras on gcp for a project with @anirban89 (see https://github.com/pangeo-data/pangeo-cloud-federation/pull/706). 

Will this cause any problems?


TomAugspurger:
I don't expect that this will cause issues as long as we already have tensorflow.

TomAugspurger:
Ah interesting. This image is explicitly installing `tensorflow-gpu>=2`. I have no idea how conda is going to handle that, since keras (via conda-forge) requires `tensorflow >=2.2` https://github.com/conda-forge/keras-feedstock/blob/331732988551fe81a5e363d1f4d076a94b8905da/recipe/meta.yaml#L25. So this might indeed cause issues. We'll see how things build.

rabernat:
Doh!

```python
from tensorflow import keras
```

This shows how old my deep learning experience is. Keras is now part of tensorflow.

PR #126, opened 2020-09-05
make base-image more generic

phaustin:
for discussion -- for our own use, I've tweaked the base-image build to:

1) add conda init information to ~joyvan/.bashrc to make the non-login and login shells symmetric wrt conda
2) add mkdir -p .../bokeh/bokeh/server/static so the find command doesn't fail


scottyhq:
@phaustin should we close this? If you could open an issue describing whatever does not work as expected with the current images that would be helpful. Thanks!

PR #131, opened 2020-09-07
add xarray_leaflet to pangeo-notebook

scottyhq:
added xarray_leaflet because it seems neat and want to test CI again


scottyhq:
can add this later in a new PR if needed

PR #143, opened 2020-10-07
Adding fusepy to environment

rsignell-usgs:
/condalock
Adding fusepy to allow trying FUSE mounts using fsspec


scottyhq:
thanks @rsignell-usgs - you no longer have to add the `/condalock` yourself in PRs, pangeo-bot takes care of it. It looks like fusepy hasn't been updated since 2016?! https://github.com/fusepy/fusepy/releases . Does it not require fuse installed with `apt.txt` ? Not sure we want to do this. Perhaps you could test with creating a separate environment in your home directory...

rsignell-usgs:
@martindurant could you please chime in here?

martindurant:
You will need FUSE installed in the system somehow, yes - I suppose this is a APT thing in this case. fusepy is very simple, because the protocol is old - I very much still expect it to work. pyfuse3 **is** better, but async takes some work. Now that (some of) fsspec is also async, it is worthwhile to use it.

PR #153, opened 2020-11-03
Add MLFlow

TomAugspurger:
I'd like to add this for the GCP hub, but I'd like to see what it does to the packages solved for /  the image size. MLFlow has a fair number of dependencies, some of which are pretty strict.

This may make more sense to just keep in the hub environment.


pangeo-notebook/conda-linux-64.lock
```diff
@@ -212,6 +216,8 @@ https://conda.anaconda.org/conda-forge/noarch/sphinxcontrib-htmlhelp-1.0.3-py_0.
 https://conda.anaconda.org/conda-forge/noarch/sphinxcontrib-jsmath-1.0.1-py_0.tar.bz2#67cd9d9c0382d37479b4d306c369a2d4
 https://conda.anaconda.org/conda-forge/noarch/sphinxcontrib-qthelp-1.0.3-py_0.tar.bz2#d01180388e6d1838c3e1ad029590aa7a
 https://conda.anaconda.org/conda-forge/noarch/sphinxcontrib-serializinghtml-1.1.4-py_0.tar.bz2#8ea6a8036e28dba8827d35c764709358
+https://conda.anaconda.org/conda-forge/linux-64/sqlalchemy-1.3.13-py38h516909a_0.tar.bz2#27c2d386486c21a4a134497abbe2d853
```
TomAugspurger, line 219: Maybe not a problem, but mlflow pins to an older version of alembic, which requires this older SQLAlchemy.

scottyhq:
> I'd like to see what it does to the packages solved for / the image size. 

@TomAugspurger for the image size piece, this makes me think we should add a `docker images` command to the github action. could you do that in this PR? should be straightforward to add an additional command or probably best as a separate step here:
 https://github.com/pangeo-data/pangeo-docker-images/blob/f7e02904549a32b8c88af3bf44fc36566bf77b65/.github/workflows/Build.yml#L78

scottyhq:
running locally I see this before:
```
REPOSITORY               TAG                 IMAGE ID            CREATED             SIZE
pangeo/ml-notebook       master              6735b9615678        20 hours ago        5.39GB
pangeo/pangeo-notebook   master              4e7c3c9c45d6        20 hours ago        3.02GB
pangeo/base-notebook     master              748b943d1168        21 hours ago        922MB
pangeo/base-image        master              d3e5e81583ce        21 hours ago        299MB
``

scottyhq (approved):

TomAugspurger:
Gonna push a couple more helpers here (papermill). Also pushed the command to list image sizes.

scottyhq:
/condalock

^ you have to manually add that in a comment if you want to resolve the environment with new packages. only runs automatically when the PR is opened.

scottyhq:
Hmm, looks like solving the environment failed now https://github.com/pangeo-data/pangeo-docker-images/runs/1349000746?check_suite_focus=true , but I don't know what to make of the error message. I'd recommend building locally to try and figure out whats happening. One option is to put this into the ml-image which is already much more complicated and uses the more lenient `channel_priority: flexible` mode in order to get tensorflow + dependencies to install:
https://github.com/pangeo-data/pangeo-docker-images/blob/master/ml-notebook/condarc.yml

TomAugspurger:
Probably from me adding a pip package.

I'll get it added to conda-forge.

scottyhq:
ah, i didn't notice. in that case, you should probably add it to requirements.txt (which is handled separately becuase conda-lock doesn't work with pip packages as far as I know). Or if you could also add `rechunker` to conda-forge and then everything could be handled through conda/mamba...
https://github.com/pangeo-data/pangeo-docker-images/blob/master/pangeo-notebook/requirements.txt

TomAugspurger:
Going to hold on this for a bit since I likely won't have time to push on MLFlow stuff in the near-term.

PR #155, opened 2020-11-06
Update workflow for deprecated set-env

TomAugspurger:
Github Actions deprecated `set-env`:
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

```
Error: The `set-env` command is deprecated and will be disabled soon.
Please upgrade to using Environment Files. For more information see:
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

The replacement is at
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files.

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -31,8 +31,8 @@ jobs:
         SHA7="${GITHUB_SHA::7}"
         DOCKER_TAG=$SHA7
         IMAGE_SPEC="${DOCKER_ORG}/${{ env.IMAGE }}:${DOCKER_TAG}"
-        echo "::set-env name=DOCKER_TAG::${DOCKER_TAG}"
-        echo "::set-env name=IMAGE_SPEC::${IMAGE_SPEC}"
+        echo "DOCKER_TAG=${DOCKER_TAG}" >> $GITHUB_ENV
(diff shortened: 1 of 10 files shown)


scottyhq:
Thanks @TomAugspurger  - one annoying aspect of our current CI is that pangeo-bot just runs /condalock automatically with each new PR. 

We should really add a path conditional so that /condalock only runs if environment.yml is changed... https://github.com/pangeo-data/pangeo-docker-images/blob/master/.github/workflows/PR.yml

Are you able to add that and revert the last bot commit to remove all the .lock and packages.txt changes which would override your last pangeo-notebook bump?

TomAugspurger:
Thanks pushed.

I'm not 100% sure if it'll take effect. The docs at https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions only mention `paths` for `push` and `pull_request`, not `pull_request_target` (which we use to get access to secrets presumably)

scottyhq:
> I'm not 100% sure if it'll take effect.

Thanks TOm! We'll see...

scottyhq (approved):

Issue #167, opened 2020-12-05
Add Jax to ML image?


rabernat:
Jax is becoming very important / useful to us. Would it be easy to add it to the ML image?

scottyhq:
> Jax is becoming very important / useful to us. Would it be easy to add it to the ML image?

Sure, what is jax used for? I'm not sure the ML image has been actively used much recently. Previously @jhamman and @jsadler2 helped decide what packages are in there. Note that it currently is also missing many of the geoscience packages that have recently been added to pangeo-notebook. But the environment solving has been complicated in the past, so generally the fewer the packages the better it seems.

In any case, to add jax, just create a PR adding a line here: 
https://github.com/pangeo-data/pangeo-docker-images/blob/7e9a10925bb5cf2574b8effd1fd9c81ecad8c6e3/ml-notebook/environment.yml#L27

rsignell-usgs:
@rabernat would be curious to know what you are going to use this for? 
I'm guessing something related to this: 
https://pymc-devs.medium.com/the-future-of-pymc3-or-theano-is-dead-long-live-theano-d8005f8a0e9b

scottyhq:
closed by #196

PR #179, opened 2021-01-16
Add pytorch and jax to ml image

rabernat:
I have no idea if this will even solve. If it does, the resulting image will probably be huge. But I want to see if it builds.


scottyhq:
one problem with the CI currently is that it's not obvious when the environment lock fails, which it did here! https://github.com/pangeo-data/pangeo-docker-images/runs/1712185623?check_suite_focus=true 

@rabernat for this fairly complicated ml-image I recommend trying some configurations following the local building instructions:
https://github.com/pangeo-data/pangeo-docker-images#how-to-build-images-locally

scottyhq:
related to #167

PR #195, opened 2021-03-18
Let base-image recognize jupyter_notebook_config.py, and configure nb_conda_kernels to not register a kernel again

consideRatio:
## PR summary

This PR solves the issue below. It does so in two steps.
1. Adds the feature base-image feature to detects a `jupyter_notebook_config.py` file and injects it to the otherwise empty folder `/etc/jupyter`.
2. It adds a `jupyter_notebook_config.py` file to the `pangeo-notebook` and `ml-notebook` folder that have `nb_conda_kernels` and configures it to avoid register a jupyter kernel twice.

## Duplicated kernel registration issue

The pangeo-notebook and ml-notebook install nb_conda_kernels, which will scan conda environments and register jupyter kernels based on installed packages in those kernels. But, if a kernel is already registered it will be registered again.

image

This can be avoided though, if we for example have a `/etc/jupyter/jupyter_notebook_config.py` available like this...

```python
import os

# Configure nb_conda_kernels to avoid registering jupyter kernels in our conda
# environment again.
c.CondaKernelSpecManager.env_filter = f'.*envs/{os.environ["CONDA_ENV"]}.*'
```

image

### DIFF
--- a/base-image/Dockerfile
+++ b/base-image/Dockerfile
@@ -134,4 +134,16 @@ ONBUILD RUN echo "Checking for 'start'..." \
         chmod +x start \
         && cp start /srv/start \
         ; fi
+
+# Copy jupyter_notebook_config.py to /etc/jupyter, which is an action not done
+# with repo2docker.
+ONBUILD USER root
+ONBUILD RUN echo "Checking for config 'jupyter_notebook_config.py'..." \
+        ; [ -d binder ] && cd binder \
+        ; [ -d .binder ] && cd .binder \
+        ; if test -f "jupyter_notebook_config.py" ; then \
+        mkdir -p /etc/jupyter \
+        && cp jupyter_notebook_config.py /etc/jupyter \
+        ; fi
+ONBUILD USER ${NB_USER}
 # ----------------------
--- a/base-image/Dockerfile
+++ b/base-image/Dockerfile
@@ -59,7 +59,8 @@ ENTRYPOINT ["/srv/start"]
 # Only run these if used as a base image
 # ----------------------
 ONBUILD USER root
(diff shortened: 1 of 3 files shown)


scottyhq:
thanks @consideRatio ! had often wondered about that, thanks for the detailed description, one minor suggestions above

base-image/Dockerfile
```diff
@@ -134,4 +134,16 @@ ONBUILD RUN echo "Checking for 'start'..." \
         chmod +x start \
         && cp start /srv/start \
         ; fi
+
+# Copy jupyter_notebook_config.py to /etc/jupyter, which is an action not done
+# with repo2docker.
+ONBUILD USER root
```
scottyhq, line 140: the first onbuild commands run as root, so if you move this section up in the Dockerfile you don't have to switch back to root for this piece

consideRatio: Ah! I put it in the bottom because it could be useful to not break the cache for the other steps, but the cache will always break because of the `COPY .` statement that proceeds all installation steps.

Since I have no clear solution for that, and it would be feature creep to solve it here, I agree and moved it to the root section!

consideRatio:
Thanks for the quick review @scottyhq! :tada: :heart:

scottyhq:
> Ah! I put it in the bottom because it could be useful to not break the cache for the other steps, but the cache will always break because of the COPY . statement that proceeds all installation steps. 

Yeah, there is probably some way to restructure things for onbuild+cache, but didn't ever dig into it since we're pretty satified with builds < 10min. But future PRs more than welcome! ;)

scottyhq (approved):

PR #207, opened 2021-05-04
add stackstac

scottyhq:
it's pretty cool https://github.com/gjoseph92/stackstac


pangeo-notebook/environment.yml
```diff
@@ -88,3 +88,5 @@ dependencies:
  - xpublish
  - xrft
  - zarr
+ - pip:
```
TomAugspurger, line 91: @scottyhq does this get picked up when you're using the lock file? Or do you need a requirements.txt?

scottyhq: whoops i totally forgot about that, you're right. might be possible eventually https://github.com/conda-incubator/conda-lock/issues/4 . But i'll follow-up to get it in there. Plus stackstac is now 0.2 :)

PR #208, opened 2021-05-05
add stackstac in pip requirements.txt

scottyhq:
following up #207 . Since we're using conda-lock files to build the image, we have a separate step for picking up pip installs from a requirements.txt instead of adding these under -pip: in the environment.yml


scottyhq:
logs now show `Successfully installed stackstac-0.2.0` , should probably modify github actions to also print out any pip package versions to avoid scrolling through all the logs...

PR #219, opened 2021-06-11
bump to python 3.9

scottyhq:
it's been out for a while now ;)

### DIFF
--- a/base-image/Dockerfile
+++ b/base-image/Dockerfile
@@ -6,7 +6,7 @@ LABEL org.opencontainers.image.source=https://github.com/pangeo-data/pangeo-dock
 # Run this section as root
 # try to keep conda version in sync with repo2docker
 # ========================
-ENV CONDA_VERSION=4.10.0-0 \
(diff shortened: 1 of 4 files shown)


scottyhq:
> Encountered problems while solving:
>  - package xlayers-0.2.1-py36he991be0_0 requires python >=3.6,<3.7.0a0, but none of the providers can be installed

Seems `xlayers` https://github.com/conda-forge/xlayers-feedstock is currently incompatible with python3.9, so I'm commenting it for now cc @cspencerjones @ocefpaf

cspencerjones:
Yes, that's definitely fine. I would recommend people use `xgcm` for this functionality now (so you could actually delete it).

ocefpaf:
> Yes, that's definitely fine. I would recommend people use `xgcm` for this functionality now (so you could actually delete it).

Do you remember the reason xlayers cannot support 3.9? Maybe it is something we can fix upstream and update the package.

PR #220, opened 2021-07-05
Added xskillscore package to docker

hdrake:
I'm using this package https://github.com/xarray-contrib/xskillscore for my cmip6hackathon projects and it would be very helpful to have in the docker image (probably to others as well).


hdrake:
It seems like this failed because of a geoviews bug (or a bug in the way it is tested).

scottyhq:
the log makes me think it is an incompatibility with pandas 1.2 -> 1.3 :
```
../conda/envs/notebook/lib/python3.8/site-packages/param/__init__.py:1406: TypeError
177
----------------------------- Captured stderr call -----------------------------
178
WARNING:param.main: pandas could not register all extension types imports failed with the following error: cannot import name 'ABCIndexClass' from 'pandas.core.dtypes.generic' (/srv/conda/envs/notebook/lib/python3.8/site-packages/pandas/core/dtypes/generic.py)
179
------------------------------ Captured log call -------------------------------
180
WARNING  param.main:parameterized.py:2009 pandas could not register all extension types imports failed with the following error: cannot import name 'ABCIndexClass' from 'pandas.core.dtypes.generic' (/srv/conda/envs/notebook/lib/python3.8/site-packages/pandas/core/dtypes/generic.py)
```

maybe try pinning `pandas~=1.2` here https://github.com/pangeo-data/pangeo-docker-images/blob/563817400983720227d0d48b0c960cf0e4d7de01/pangeo-notebook/environment.yml#L53

hdrake:
> maybe try pinning `pandas~=1.2` here

Thanks @scottyhq – I'll see if that does it.

hdrake:
> the log makes me think it is an incompatibility with pandas 1.2 -> 1.3 :
> 
> ```
> ../conda/envs/notebook/lib/python3.8/site-packages/param/__init__.py:1406: TypeError


Maybe I'm reading the error wrong, but isn't everything below that just a warning and the actual error this part above?

```
self = <param.List object at 0x7fc54b1ed460>, val = None, bounds = (0, None)

    def _validate_bounds(self, val, bounds):
        "Checks that the list is of the right length and has the right contents."
        if bounds is None:
            return
        min_length, max_length = bounds
>       l = len(val)
E       TypeError: object of type 'NoneType' has no len()

../conda/envs/notebook/lib/python3.8/site-packages/param/__init__.py:1406: TypeError
```

scottyhq:
Yeah. Sorry  @hdrake to be a bother, but a better solution is probably to pin `param=1.10` and go back to unpin pandas, could you please push that change, if it works we'll merge that, otherwise we can temporarily restrict pandas.

i didn't see any issues currently open in https://github.com/holoviz/param , so would be good to add something there with a more minimal environment that reproduces the error.

hdrake:
@scottyhq, it actually looks like they know of this issue and fixed it between param=1.11.0 and param=1.11.1, at least based on the bug-fix mentioned in the release notes: https://github.com/holoviz/param/releases/tag/v1.11.1.

I'll try just excluding 1.11.0 instead.

scottyhq (approved):

Issue #248, opened 2021-09-24
`requirements.txt` `stackstac` overwrites the `rasterio` conda package with a pip package. This eliminates the support for HDF4 format.


BnJam:
Conda installs `rasterio-1.1.8` which is the full binary with HDF4 support. Pip installs `stackstac` which has a dependency for `rasterio==1.2.6` that overwrites the Conda package and replaces it with the pip package elimating the HDF4 format support. 

Tags affects: 2021.05.15 and onward.

TomAugspurger:
Looks like stackstac is in conda-forge now: https://github.com/conda-forge/stackstac-feedstock

Would you be able to make a PR moving it from requirements.txt to the environment.yaml?

PR #249, opened 2021-09-25
#248 removing stackstac from requirements into environment.yml as it'…

BnJam:
…s available in conda-forge

Moved `stackstac` from requirements to environment.yml. 

When `stackstac` is in requirements.txt, it is installed via Pip and the conda-forge installed `rasterio-1.2.8` is eliminated in place of PyPi `rasterio==1.2.6`. The PyPi installation does not have HDF4 format support. Now `stackstac-0.2.1` is available on conda-forge, the requirements.txt is no longer needed.


scottyhq:
thanks @BnJam ! the `packages.txt` summaries get added automatically by CI, so for future PRs you actually don't need to include those

scottyhq:
/condalock

re-running ...

pangeo-notebook/packages.txt
```diff
@@ -135,22 +135,22 @@ freeglut-3.2.1
...
 geocube-0.0.17
 geographiclib-1.52
 geopandas-0.9.0
 geopandas-base-0.9.0
 geopy-2.2.0
 geos-3.9.1
-geotiff-1.6.0
-geoviews-core-1.9.1
+geotiff-1.7.0
+geoviews-core-1.8.2
```
scottyhq, line 153: not sure why `geoviews-core` dropped from 1.9.1 to 1.8.2... also not sure if we need to keep using `geoviews-core` or just geoviews...

scottyhq (approved):

BnJam:
> thanks @BnJam ! the `packages.txt` summaries get added automatically by CI, so for future PRs you actually don't need to include those

Thanks @scottyhq ! 

I saw that the packages.txt isn't being used by any of the Dockerfiles, but I was uncertain about not explicitly calling a version number from the environmental.yml file.

PR #256, opened 2021-10-18
Add RISE and ipytree to pangeo-notebook

rabernat:
Instead of https://github.com/conda-forge/pangeo-notebook-feedstock/pull/86

Also noting that xlayers is not needed as its functionality is better supported in xgcm.


rabernat:
> you can use the binder links in the comment to test

This is such an amazing feature! 🎉 

The RISE button is now available via the legacy notebook:

image

Zarr groups now show HTML reprs:

image

Those were my main goals with this PR

PR #265, opened 2021-11-04
export conda-env-path in requirements.txt and start build steps

jhamman:
This is trying to fix an issue where executables in the conda environment are not in the environments PATH during the `requirements.txt` and `start` build steps. I ran into this when installing `git` via conda and installing a python package from source via the `requirements.txt` file.

### DIFF
--- a/base-image/Dockerfile
+++ b/base-image/Dockerfile
@@ -17,8 +17,9 @@ ENV CONDA_VERSION=4.10.3-6 \
 
 ENV NB_PYTHON_PREFIX=${CONDA_DIR}/envs/${CONDA_ENV} \
     DASK_ROOT_CONFIG=${CONDA_DIR}/etc \
-    HOME=/home/${NB_USER} \
-    PATH=${CONDA_DIR}/bin:${PATH}
+    HOME=/home/${NB_USER}
+    
+ENV PATH=${NB_PYTHON_PREFIX}/bin:${CONDA_DIR}/bin:${PATH}
 
 # required for prefect agent: https://github.com/PrefectHQ/prefect/issues/3061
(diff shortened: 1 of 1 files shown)


base-image/Dockerfile
```diff
@@ -121,7 +121,8 @@ ONBUILD RUN echo "Checking for pip 'requirements.txt'..." \
         ; [ -d binder ] && cd binder \
         ; [ -d .binder ] && cd .binder \
         ; if test -f "requirements.txt" ; then \
-        ${NB_PYTHON_PREFIX}/bin/pip install --no-cache-dir -r requirements.txt \
+        export PATH=${NB_PYTHON_PREFIX}/bin:${PATH} \
```
scottyhq, line 124: @jhamman can you just add this to the ENV block so that it's in one place? https://github.com/pangeo-data/pangeo-docker-images/blob/690362786bc1da03e2cb953982a14d64a91815b4/base-image/Dockerfile#L18

base-image/Dockerfile
```diff
@@ -121,15 +122,14 @@ ONBUILD RUN echo "Checking for pip 'requirements.txt'..." \
         ; [ -d binder ] && cd binder \
         ; [ -d .binder ] && cd .binder \
         ; if test -f "requirements.txt" ; then \
-        ${NB_PYTHON_PREFIX}/bin/pip install --no-cache-dir -r requirements.txt \
+        && ${NB_PYTHON_PREFIX}/bin/pip install --no-cache-dir -r requirements.txt \
```
scottyhq, line 125: remove && coming after `then` (similarly for changes below)

jhamman: 🤦

scottyhq (approved):

PR #276, opened 2021-12-07
Add conda create command with remote lockfile

scottyhq:
This has come up in a few places, but it's worth documenting in the readme I think.

The idea is that sometimes people just want the conda environment (e.g. not actually using a dask cluster, or running on a hosted service like Colab or Sagemaker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ If you're used to managing conda environments on your personal computer, or runn
 ```
 %conda create -n pangeo-notebook --file https://raw.githubusercontent.com/pangeo-data/pangeo-docker-images/2021.12.02/pangeo-notebook/conda-linux-64.lock
 ```
+Note that this will only work on linux environments, since the conda lockfile is specific to linux.
 
 ### Image tagging and "continuous building"
 This repository uses [GitHub Actions](https://help.github.com/en/actions) to build images, run tests, and push images to [DockerHub](https://hub.docker.com/orgs/pangeo).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,6 +68,14 @@ docker run -it --rm pangeo/base-notebook:2021.09.30 /bin/bash
 Many Cloud providers offer services to run Docker containers in their data centers. Instructions will vary, so we don't provide specifics here, but as an example you can check out these docs for running containers on [AWS ECS via Docker Compose](https://docs.docker.com/cloud/ecs-integration/)
 
 
+### How to install just the conda environment
+
+If you're used to managing conda environments on your personal computer, or running a hosted JupyterLab service like [Google Colab](https://colab.research.google.com) or [AWS SageMaker Studio Lab](https://studiolab.sagemaker.aws), you can exactly match a tagged pangeo-notebook conda environment. For example, below we install the `pangeo-notebook` environment tagged on `2021.12.02`:
+
+```
+%conda create -n pangeo-notebook --file https://raw.githubusercontent.com/pangeo-data/pangeo-docker-images/2021.12.02/pangeo-notebook/conda-linux-64.lock
```
jhamman, line 76: it may be worth also exporting lock files for other platforms. Our team [uses](https://github.com/carbonplan/envs/blob/cb175a89972c6d2e3b6e8a740626359e512f6709/cmip6-downscaling/02_make_lock_files.sh#L10) this method for linux and osx environments, both of which can be exported from `conda-lock`.

README.md
```diff
@@ -68,6 +68,14 @@ docker run -it --rm pangeo/base-notebook:2021.09.30 /bin/bash
...
+```
+
```
rabernat, line 78:
```suggestion
Note that this will only work on linux environments, since the conda lockfile is specific to linux.

```

README.md
scottyhq: i thought we could just stick to linux for simplicity, but could add these later if need be.

PR #282, opened 2022-01-26
Replace version pinned Miniforge with latest Mambaforge

consideRatio:
A maintenance update of the miniforge version and mamba version. To avoid needing to make this kind of update too often, I've also added logic to specify the miniforge version by `major.minor` instead of doing it with `major.minor.patch-build`.

- The miniforge version was pinned to `4.10.3-7` and is via this PR pinned to be `4.11` which currently resolves to `4.11.0-0`.
- The mamba version was pinned to `0.17` and is now pinned to `0.20` which currently resolves to `0.20.0`.

Note that to specify the miniforge version by `major.minor` only, we needed logic to find the latest version matching that as the installation process is to download a specific URL including the exact version. This logic was implemented using a GitHub API and the CLI tool `jq` which was also added to parse out the latest matching version.
 
### Related issue

This PR is a result of me trying to solve https://github.com/pangeo-data/jupyter-earth/issues/101 by updating the mamba version where we rely on the `pangeo/pangeo-notebook` image as a base image.

### DIFF
--- a/base-image/Dockerfile
+++ b/base-image/Dockerfile
@@ -4,10 +4,14 @@ FROM ubuntu:20.04
 LABEL org.opencontainers.image.source=https://github.com/pangeo-data/pangeo-docker-images
 
 # Run this section as root
-# try to keep conda version in sync with repo2docker?
-# https://github.com/jupyterhub/repo2docker/blob/297c7955d208f3595c4d5bc2b079953199bb1f35/repo2docker/buildpacks/conda/install-miniforge.bash#L8-L9# ========================
-ENV CONDA_VERSION=4.10.3-7 \
-    MAMBA_VERSION=0.17 \
+# FIXME: Pin CONDA_VERSION only to "major.minor". Currently we can't just set
+#        CONDA_VERSION without including ".patch-build" as its used as a part in
+#        a download URL that requires the full version specification. The github
+#        repo's tags are available before the GitHub releases are made as well.
+# CONDA_VERSION: A miniforge release version https://github.com/conda-forge/miniforge/releases
+# MAMBA_VERSION: A conda-forge package version https://anaconda.org/conda-forge/mamba
+ENV CONDA_VERSION=4.11.0-0 \
+    MAMBA_VERSION=0.20 \
     CONDA_ENV=notebook \
     NB_USER=jovyan \
     NB_UID=1000 \
--- a/base-image/Dockerfile
+++ b/base-image/Dockerfile
@@ -4,13 +4,13 @@ FROM ubuntu:20.04
 LABEL org.opencontainers.image.source=https://github.com/pangeo-data/pangeo-docker-images
 
 # Run this section as root
-# FIXME: Pin CONDA_VERSION only to "major.minor". Currently we can't just set
-#        CONDA_VERSION without including ".patch-build" as its used as a part in
-#        a download URL that requires the full version specification. The github
-#        repo's tags are available before the GitHub releases are made as well.
-# CONDA_VERSION: A miniforge release version https://github.com/conda-forge/miniforge/releases
-# MAMBA_VERSION: A conda-forge package version https://anaconda.org/conda-forge/mamba
-ENV CONDA_VERSION=4.11.0-0 \
+# CONDA_VERSION: A miniforge release version found at
+#                https://github.com/conda-forge/miniforge/releases specified to
+#                either just "major" or "major.minor", but always excluding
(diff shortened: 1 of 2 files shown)


base-image/Dockerfile
```diff
@@ -4,10 +4,14 @@ FROM ubuntu:20.04
...
-ENV CONDA_VERSION=4.10.3-7 \
-    MAMBA_VERSION=0.17 \
+# CONDA_VERSION: A miniforge release version found at
+#                https://github.com/conda-forge/miniforge/releases specified to
+#                either just "major" or "major.minor", but always excluding
+#                ".patch-build".
+# MAMBA_VERSION: A conda-forge package version found at
+#                https://anaconda.org/conda-forge/mamba
+ENV CONDA_VERSION=4.11 \
+    MAMBA_VERSION=0.20 \
```
scottyhq, line 14: @consideRatio I'm curious if you need mamba=0.20, an alternative simplification could be to just use CONDA_VERSION and pull the newer "mambaforge" installers that did not previously exist (https://github.com/conda-forge/miniforge/releases/download/4.11.0-0/Mambaforge-4.11.0-0-Linux-x86_64.sh). It's not clear to me what mamba version is packaged within though...

consideRatio: Going straight for a mambaforge installer makes perfect sense to me.

I can do that as part of this PR if you think such change is okay. Doing so, would it be okay if i removed the environment variables CONDA_VERSION / MAMBA_VERSION as well and added MAMBAFORGE_VERSION?

scottyhq: yes sounds great! as far as I can tell, it's the same setup just with mamba included by default

consideRatio: Going for it!

scottyhq:
FYI - the miniforge version is currently specified in a couple places, because of how we generate a lockfile for every PR via github actions. https://github.com/pangeo-data/pangeo-docker-images/blob/9b282f51eb52000dfba40f8aad972532fba5e07f/.github/workflows/CondaLock.yml#L37-L44 

I think a better strategy going forward might be to install `conda-lock` in the base image Dockerfile and then actually run conda-lock via that Docker image. That way, conda, mamba, and conda-lock versions will be consistent throughout this repository... 

But happy to merge this first and follow up separately with changing the github actions!

base-image/Dockerfile
consideRatio:
```
    && conda install -y -c conda-forge mamba=${MAMBA_VERSION} \
    && mamba clean -afy \
    && find ${CONDA_DIR} -follow -type f -name '*.a' -delete \
    && find ${CONDA_DIR} -follow -type f -name '*.pyc' -delete
```

Do you have ideas on if this stuff needs to be cleaned up still?

scottyhq: yes, i think all 4 lines can be removed now

consideRatio: I've concluded those cleanup commands still has work to do, at least the last two.

consideRatio:
@scottyhq hmmm about this reference that should be kept in sync...

https://github.com/pangeo-data/pangeo-docker-images/blob/9b282f51eb52000dfba40f8aad972532fba5e07f/.github/workflows/CondaLock.yml#L37-L44 

Following the logic on how the github action's `miniforge-version` parameter is used to [this](https://github.com/conda-incubator/setup-miniconda/blob/723ab12c1d66bb893c5ac701f8e387f00b24549e/src/installer/download-miniforge.ts#L31-L43), I conclude the option is to either pin the version fully or use `latest`.

I conclude that I can't make them match well, unless we go for `latest` systematically or full pinning systematically. I've grown quite fond of not pinning version when maintaining jupyter/docker-stacks over the last year. What do you think about removing the pinning and going with latest?

base-image/Dockerfile
scottyhq: interesting! thanks for checking

scottyhq:
> What do you think about removing the pinning and going with latest? 

Sounds good to me.

> I didn't understand this =/ I don't overview the various automation parts etc in this repo. 

No worries, sorry if I confused you, I might ask for your review in a follow-up PR where I'd be able to explain in more detail.

consideRatio:
@scottyhq okay ready for review again I think!

scottyhq (approved):
thanks @consideRatio !

consideRatio:
Thanks @scottyhq for your quick review/merge!

PR #291, opened 2022-03-01
Add argopy to pangeo-notebook env

gmaze:
Add the [argopy](https://github.com/euroargodev/argopy) package


gmaze:
Hi,
I don't see why the ml-notebook test fails
But I found that it tried to install the argopy version 0.1.6 while we're in 0.1.9m with lots of changes in requirements

I also could install argopy 0.1.9 and build image with pangeo/ml-notebook:2022.02.04

Example: [Binder](https://mybinder.org/v2/gh/euroargodev/binder-sandbox/pangeo-ml?urlpath=git-pull%3Frepo%3Dhttps%253A%252F%252Fgithub.com%252Feuroargodev%252FOSnet-GulfStream%26urlpath%3Dlab%252Ftree%252FOSnet-GulfStream%252Fdocs%252Fdemo-predictions.ipynb%26branch%3Dbinder)

pangeo-notebook/conda-linux-64.lock
```diff
@@ -611,3 +614,7 @@ https://conda.anaconda.org/conda-forge/noarch/ipyurl-0.1.2-pyh3684270_1.tar.bz2#
 https://conda.anaconda.org/conda-forge/noarch/jupyterhub-singleuser-2.1.1-hd8ed1ab_0.tar.bz2#86aab26e793c1527e5240dfb08fc08aa
 https://conda.anaconda.org/conda-forge/noarch/pangeo-notebook-2022.01.31-hd8ed1ab_0.tar.bz2#ccd7fecfc6a66582b7269b3f1940d883
 https://conda.anaconda.org/conda-forge/noarch/xarray_leaflet-0.1.15-pyhd8ed1ab_0.tar.bz2#33925089afaae43154afe188bdbe0670
+https://conda.anaconda.org/conda-forge/noarch/argopy-0.1.6-pyhd8ed1ab_1.tar.bz2#54e0baf97064216935453ee380ec520d
```
rabernat, line 617: 0.1.6 is also getting installed in the regular pangeo-notebook image. I'm not sure why. Perhaps you want to pin a version >= 0.1.9. Although I don't understand why it doesn't pick the latest version automatically.

rabernat:
It looks like the condalock command is failing. https://github.com/pangeo-data/pangeo-docker-images/actions/runs/1922164853

At this point we probably need some help from @scottyhq, who is on US west coast time.

Thanks for your patience Guillaume! We will get this working! Very excited to have argopy added to this image.

gmaze:
Great ! No worries, no urgencies
Thanks for your help !

scottyhq:
@gmaze I notice argopy has fairly strict dependency pins: https://github.com/euroargodev/argopy/blob/master/requirements.txt . The upper version pin in particular can be hard to solve for with the many other packages already in the environment. 

As an example, `xarray>=0.15.1, <0.21.2` will not allow for the latest xarray version (which just switched to 'calendar' style versioning https://github.com/pydata/xarray/releases/tag/v2022.03.0)

gmaze:
> argopy has fairly strict dependency pins

Hi @scottyhq , indeed we had to impose this last year because we couldn't catch up with breaking changes associated with new releases. 

Do you expect argopy package to:
- [ ] do not have upper bounds on dependency pins ?
- [ ] allow upper bounds as long as it's the last version ? (which I guess will fall back on the previous case)


We implemented CI tests with last versions of dependencies last year in order to allow us to keep argopy compatible with the these most recent versions
But the monthly frequency of new releases (especially for xarray and fsspec) is not always compatible with our man power (which basically nails down to me most of the time 😢 ) to keep up

I guess I could find a way to make automatic monthly release of argopy as long as CI tests passed with last versions. But I'm not confortable with creating new releases without more than a requirements update.
 
If you guys have ideas or suggestions about this, I'll appreciate

rabernat:
> But I'm not confortable with creating new releases without more than a requirements update.

Could you explain this a bit more? What is the downside to making such releases?

gmaze:
> Could you explain this a bit more? What is the downside to making such releases?

There is absolutely nothing new in the software that would be "improve" user experience.
But now that I'm writing it, I realise that keeping the software compatible with an ever evolving environments would improve the user experience obviously ! 😜 

Then I guess I was looking for an excuse to stay away from one more complex machinery (automatic monthly release) that would drain my energy, 🔋  I would prefer to use to develop new features.

rabernat:
I see where you are coming from!

In our projects, we try to make releasing as easy as possible using automation, so there is as little burden on the maintainers to release often. There is really no downside (other than this burden) for releasing a lot. For example, here is a [workflow](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/master/.github/workflows/release.yaml) that creates a new pypi release whenever you create a release on GitHub. So we just have to click a button to make the release (see [developer docs](https://pangeo-forge.readthedocs.io/en/latest/development/development_guide.html#releasing)).

However, the main question is _whether argopy is actually compatible with the latest version of xarray_. In general, packages like xarray should avoid making changes in a way that break dependent packages such as argopy. What happens if you just remove the upper bound on xarray? Do the tests fail with xarray==2022.03.0?

rabernat:
In https://github.com/euroargodev/argopy/pull/182 I am proposing that you just remove the upper version bound from all your dependencies. I think this will fix all of the problems here.

After that you can manually make a new release.

After making this change, I don't think automatically releasing every month, regardless of whether there have been changes, would be necessary.

gmaze:
Ok, all CI tests passed after https://github.com/euroargodev/argopy/pull/182
I'll draft a new release tomorrow and will keep you posted !
Thanks a lot @rabernat

pangeo-notebook/environment.yml
```diff
@@ -6,6 +6,7 @@ dependencies:
  - pangeo-notebook=2022.01.31
  - pip=20
  - adlfs
+ - argopy >= 0.1.9
```
rabernat, line 9:
```suggestion
 - argopy
```

Once the new argopy release is up on conda forge, let's go back to this.

gmaze:
New release is out on [pypi](https://pypi.org/project/argopy/#history), waiting for [conda-forge feedstock](https://github.com/conda-forge/argopy-feedstock/pulls)  to catch up

gmaze:
hi @rabernat 
Argopy has moved to 0.1.10 on Conda, removing upper pins on requirements 
May be we can give this another try ...
Thanks

gmaze:
> This has gone a little stale somehow

A 1 week old PR is stalled ? we don't live on the same time axis 🚀  😆

PR #292, opened 2022-03-02
Document the base-image & minor cleanup

yuvipanda:
- Add a lot of inline documentation to the base-image, which is where
  almost all the magic is.
- Install base-image packages in one go, rather than in two
  lines - this is slightly faster.
- Hide stdout from apt-get install from base-image and from
  apt-get update, but not for packages from apt.txt. Earlier
  stderr was being hidden - but that is useful when the package
  installation actually fails
- Use `--no-cache` instead of `--no-cache-dir` when installing
   with pip, to be clearer on what we want.

### DIFF
--- a/base-image/Dockerfile
+++ b/base-image/Dockerfile
@@ -4,34 +4,46 @@ FROM ubuntu:20.04
 
 LABEL org.opencontainers.image.source=https://github.com/pangeo-data/pangeo-docker-images
 
-# Run this section as root
+# Setup environment to match variables set by repo2docker as much as possible
+# The name of the conda environment into which the requested packages are installed
 ENV CONDA_ENV=notebook \
+    # Tell apt-get to not block installs by asking for interactive human input
+    DEBIAN_FRONTEND=noninteractive \
+    # Set username, uid and gid (same as uid) of non-root user the container will be run as
     NB_USER=jovyan \
     NB_UID=1000 \
+    # Use /bin/bash as shell, not the default /bin/sh (arrow keys, etc don't work then)
     SHELL=/bin/bash \
+    # Setup locale to be UTF-8, avoiding gnarly hard to debug encoding errors
     LANG=C.UTF-8  \
     LC_ALL=C.UTF-8 \
-    CONDA_DIR=/srv/conda
+    # Install conda in the same place repo2docker does
+    CONDA_DIR=/srv/conda \
+    # Path to the python environment where the jupyter notebook packages are installed
+    NB_PYTHON_PREFIX=${CONDA_DIR}/envs/${CONDA_ENV} \
+    # Home directory of our non-root user
+    HOME=/home/${NB_USER} \
+    # Add both our notebook env as well as default conda installation to $PATH
+    PATH=${NB_PYTHON_PREFIX}/bin:${CONDA_DIR}/bin:${PATH}
 
-ENV NB_PYTHON_PREFIX=${CONDA_DIR}/envs/${CONDA_ENV} \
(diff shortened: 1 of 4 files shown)


base-image/Dockerfile
```diff
@@ -1,67 +1,95 @@
...
+    # Create non-root user, with given gid, uid and create $HOME
     && useradd --create-home --gid ${NB_UID} --no-log-init --uid ${NB_UID} ${NB_USER} \
-    && echo ". ${CONDA_DIR}/etc/profile.d/conda.sh ; conda activate ${CONDA_ENV}" > /etc/profile.d/init_conda.sh \
+    # Make sure that /srv is owned by non-root user, so we can install things there
     && chown -R ${NB_USER}:${NB_USER} /srv
 
-# SEE: https://github.com/phusion/baseimage-docker/issues/58
-ARG DEBIAN_FRONTEND=noninteractive
+# Run conda activate each time a bash shell starts, so users don't have to manually type conda activate
+# Note this is only read by shell, not by python
```
consideRatio, line 45: The "not by python" part, does it relate to something like "if you start a process directly via python, and not via a shell, the environment will not be activated?"

Hmmm, can you clarify this comment? I think it contains a very relevant understanding!!

base-image/Dockerfile
```diff
@@ -1,67 +1,95 @@
...
     && find ${CONDA_DIR} -follow -type f -name '*.pyc' -delete
 
 EXPOSE 8888
 ENTRYPOINT ["/srv/start"]
 #CMD ["jupyter", "notebook", "--ip", "0.0.0.0"]
 
-# Only run these if used as a base image
+# We use ONBUILD (https://docs.docker.com/engine/reference/builder/#onbuild)
+# to support triggering certain behavior when specific files exist in the directories of our
+# child images (such as base-notebook, pangeo-notebook, etc). This is
```
consideRatio, line 77: Can you clarify when it triggers? Okay "on build", but a confusion point for me that was that ONBUILD can remain relevant even if a subsequent image is built that derives from this image as a base image. Perhaps you could try include that aspect - that the ONBUILD statement here remain relevant later as well - or not - I never remember how ONBUILD works!

base-notebook/Dockerfile
```diff
@@ -1 +1,5 @@
+# ONBUILD instructions in base-image/Dockerfile are used to
+# perform certain actions based on the presence of specific
+# files (such as conda-linux-64.lock, start) in this repo.
+# Refer to the base-image/Dockerfile for documentation.
```
consideRatio, line 4: Ah seeing this makes another comment above clear.

consideRatio (approved):
Great work with this @yuvipanda, these comments help anyone understand whats going on far quicker than without them!

This LGTM for what its worth, I left a comment on a detail to potentially clarify but I think this is great as it is already as well!

base-image/Dockerfile
yuvipanda: @consideRatio i added an example, not sure if that makes it clearer or more confusing!

base-image/Dockerfile
yuvipanda: @consideRatio added a comment, let me know if this helps!

yuvipanda:
Documentation added as part of exploration and understanding in https://www.youtube.com/watch?v=f-Q3muKzKM4. I'm going to try to add more 'overview' documentation in a separate PR later.

scottyhq (approved):
Thank you so much @yuvipanda, this additional documentation is fantastic, and the youtube recording is really nice for even more context!

PR #293, opened 2022-03-03
Use a python script to generate package list

yuvipanda:
- Lets us put a line up front in packages.txt about
  where this file is coming from, and why it exists!
  Without that, it was a bit confusing, as packages.txt is
  a secondary source, not a primary one
- Provide packages.txt in format produced by pip freeze -
  earlier it was just in a format I haven't seen anywhere
  else. This does make diffs a bit more complex before
  and after this change, but I think it is worth it to be
  clearer.


yuvipanda:
Part of live streamed exploration of https://www.youtube.com/watch?v=f-Q3muKzKM4

Makefile
```diff
@@ -12,22 +12,22 @@ base-image :
 base-notebook : base-image
 	cd base-notebook ; \
 	conda-lock lock --mamba -f environment.yml -p linux-64; \
-	../list_packages.sh | sort > packages.txt; \
+	../generate-packages-list.py conda-linux-64.lock | sort > packages.txt; \
```
scottyhq, line 15:
```suggestion
	../generate-packages-list.py conda-linux-64.lock > packages.txt; \
```

Makefile
```diff
@@ -12,22 +12,22 @@ base-image :
...
+	../generate-packages-list.py conda-linux-64.lock > packages.txt; \
 	docker build -t pangeo/pangeo-notebook:master . ; \
 	docker run -w $(TESTDIR) -v $(PWD):$(TESTDIR) pangeo/pangeo-notebook:master ./run_tests.sh pangeo-notebook
 
 .PHONY: ml-notebook
 ml-notebook : base-image
 	cd ml-notebook ; \
 	conda-lock lock --mamba -f environment.yml -f ../pangeo-notebook/environment.yml -p linux-64; \
-	../list_packages.sh | sort > packages.txt; \
+	../generate-packages-list.py conda-linux-64.lock | sort > packages.txt; \
```
scottyhq, line 31:
```suggestion
	../generate-packages-list.py conda-linux-64.lock > packages.txt; \
```

generate-packages-list.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python3
+"""
+Generate human readable packages.txt file from a conda-lock file.
+
+Looks for a conda-lock
```
scottyhq, line 5:
```suggestion
```

scottyhq:
> So instead of adding more hackery to the bash file, I rewrote it in python.

Awesome :) There has been a fair amount of bash hackery in this repository

> This does make diffs a bit more complex before and after this change, but I think it is worth it to be clearer. 

Agreed! I think the diffs are most helpful between subsequent images anyhow, so going forward things will look just fine.

yuvipanda:
@scottyhq ty for the catches! I've fixed the suggestions :)

scottyhq (approved):

PR #296, opened 2022-03-09
Add argopy to notebook env

rabernat:
(supersedes #291)


rabernat:
Let this go stale again. 😢

PR #311, opened 2022-04-26
Set nvidia specific env vars in ml-notebook

yuvipanda:
These are required for `nvidia-smi` and generally anything that touches the GPU to
properly work on most containerized systems. We could set this at the hub level, but
this is cleaner and reduces workload for hub admins.

### DIFF
--- a/ml-notebook/Dockerfile
+++ b/ml-notebook/Dockerfile
@@ -3,3 +3,7 @@
 # files (such as conda-linux-64.lock, start) in this repo.
 # Refer to the base-image/Dockerfile for documentation.
 FROM pangeo/base-image:master
+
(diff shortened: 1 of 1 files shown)


yuvipanda:
Aaa, this is only needed on GKE - Azure doesn't setup this non-standard path. You think they'd all agree to do one thing. 

I'm going to close this, as unfortunately it's not platform independent :(

yuvipanda:
Actually this will be *very* helpful on GKE, as $PATH is hard to easily modify from z2jh config - we need to preserve the current value set in the container image and augment, which is a bit difficult.

rabernat (approved):
Thanks so much Yuvi!

Issue #188, opened 2021-02-12
ML image update


dhruvbalwada:
Was talking to @scottyhq about using the ML image over here and having pytorch preloaded. I know @rabernat has asked about this before (#179) . 

We were wondering who all are using the ML image? and what might be the requirements they have? @nbren12 @jhamman 
It seems like the usage for the ML image is low based on the pulls here: https://github.com/pangeo-data/pangeo-docker-images. 

Since pytorch and tensorflow are two of the big candidates,(and maybe used independently usually), @scottyhq suggested having a pangeo-pytorch and a pangeo-tensorflow. 

Any other thoughts that people have?

rabernat:
Correct, we are not using them much right now. However, there are several project spinning up now that will require ML Pangeo images, so it's a good time to think about this.

IMO, before creating more images, we need to make a plan to address how to maintain these images sustainably going forward. Within a month or so we should have a dedicated, full-time Pangeo engineer at 2i2c, and that person should be able to help out with this.

nbren12:
I don’t use these images.

My $0.02: the many images problem is a symptom of a docker container not being a package manager. Dockerfiles are a linear sequence of commands while packages form a dependency graph. It will always be hard to map docker containers onto the packages people want.

Maintaining multiple images is painful. Honestly, for scientific workflows with GB/TB scale datasets, “light” containers don’t seem worth the trouble. If you can get away with it, I suggest 1 mega docker image (you need to pin of all package versions or it will constantly break) or leveraging a tool like repo2docker if you need multiple images. You can also e.g. have packages installed when a user starts a container like the dask image does.

nbren12:
It looks like this repo already uses repo2docker, so maybe the tooling is good enough to support many images 🤷 . Maybe pin the “from <blah>-image” statements as well to keep things more reproducible.

scottyhq:
@nbren12 thanks for the comments. This repo is a bit confusing to understand, despite the tags images are in theory reproducible thanks to using conda-lock to presolve for the environment added to the docker image, so for example to recreate an image from the past:
```
git clone https://github.com/pangeo-data/pangeo-docker-images.git
cd pangeo-docker-images
git checkout 2020.09.30
docker build -t pangeo/base-image:master base-image
docker build -t pangeo/ml-notebook:2020.09.30 ml-notebook
```

[...]

PR #315, opened 2022-04-27
Create pytorch-notebook docker image

### LINKED ISSUE #312: Add pytorch to ML image (or create separate image)
For LEAP and M2LInES 2i2c Hubs, many researchers will want to use pytorch. Unfortunately, this image does not have pytorch.

When I try to add pytorch-gpu to `ml-notebook/environment.yml` and call

```
conda-lock lock --mamba -f environment.yml -f ../pangeo-notebook/environment.yml -p linux-64
```

I get the error

```
Encountered problems while solving.
Problem: nothing provides __glibc >=2.17,<3.0.a0 needed by pytorch-1.10.0-cuda102py37h689c94d_1
```

I found a similar problem in https://github.com/NVIDIA/MinkowskiEngine/issues/354. That issue implies we may need to mix conda channels to get this to work. The [workaround](https://github.com/NVIDIA/MinkowskiEngine/issues/354#issuecomment-880555195) they recommend is to install openblas from `anaconda` and then switch to conda-forge for pytorch.

Any thoughts on the best path here?

cc @yuvipanda  

Also related to #188

weiji14:
Another GPU-enabled docker image for running deep learning!

List of packages to include (let me know of any other relevant ones);
- [x] pytorch
- [x] torchgeo
- [ ] pytorch-lightning
- [ ] etc

Resolves #312


pangeo-notebook/environment.yml
```diff
@@ -3,6 +3,9 @@ channels:
  - conda-forge
 dependencies:
  - python=3.9*
+ # needed for pytorch to work
+ # https://github.com/NVIDIA/MinkowskiEngine/issues/354#issuecomment-880555195
+ - anaconda::libopenblas
```
scottyhq, line 8: let's move this to pytorch-notebook/environment.yml

weiji14: Ok, done at 536d580f3f09e45a9a690294f437b4f6b5a3b55a

scottyhq:
Thanks for this @weiji14 ! If you add new packages (torchgeo) you have to relock the environment. The easiest way to do this is add a comment with `/condalock` on the first line.

weiji14:
> Thanks for this @weiji14 ! If you add new packages (torchgeo) you have to relock the environment. The easiest way to do this is add a comment with `/condalock` on the first line.

Ah ok, I've actually ran `conda-lock` locally already so there might not be stuff to update. But there might be a few other packages we might add so will do that later :smile:

pytorch-notebook/environment.yml
```diff
@@ -0,0 +1,12 @@
+# Note: this environment will be merged with the pangeo-notebook/environment.yml
+# by conda-lock. Only put things here that are not already in pangeo-notebook
+name: pangeo
+channels:
+ - conda-forge
+dependencies:
+ - cudatoolkit=11
```
scottyhq, line 7: in the ml-notebook image we pin cudatoolkit=10, https://github.com/pangeo-data/pangeo-docker-images/blob/b5c38d09ca44e66dfda529fbfeb15ba4789a3805/ml-notebook/environment.yml#L7

I'm guessing if both these images are offered on pangeo jupyterhubs we want cuda versions to be the same, @yuvipanda any recommendations for 10 vs 11 when it comes to node configuration? 

Fine leaving it as 11 here and bumping the other image in a separate PR.

weiji14: Yes, it comes down to what CUDA drivers the server is using (find out using `nvidia-smi`). I prefer `cudatoolkit=11` because it has foward and backward compatibility within a minor version (see https://docs.nvidia.com/deploy/cuda-compatibility/#forward-compatibility-title), but that's of no use if the cuda drivers on the server don't support it.

yuvipanda: I don't actually know enough to have an opinion here, but happy to stick to whatever upstream recommends and as long as we are consistent across the images :D If we need to upgrade the driver version on the clusters when we bump this, can do that.

weiji14: Ok. Just to confirm, I'm assuming that the clusters don't pull the latest pangeo-docker image? I.e. that they're pinned to a specific pangeo-docker-image version? Probably doesn't matter for this `pytorch-notebook` image since nobody will be using it yet, but I'd recommend updating the CUDA driver version before bumping the `cudatoolkit` version on `ml/tensorflow-notebook`, just speaking from experience working on an on-prem HPC :slightly_smiling_face:

README.md
```diff
@@ -13,6 +13,7 @@ Image are hosted on DockerHub: https://hub.docker.com/u/pangeo
 | base-image      | Foundational Dockerfile for builds            | ![](https://img.shields.io/docker/image-size/pangeo/base-image?sort=date) | ![](https://img.shields.io/docker/pulls/pangeo/base-image?sort=date)
 | [base-notebook](base-notebook/packages.txt) | minimally functional image for pangeo hubs | ![](https://img.shields.io/docker/image-size/pangeo/base-notebook?sort=date) | ![](https://img.shields.io/docker/pulls/pangeo/base-notebook?sort=date)
 | [pangeo-notebook](pangeo-notebook/packages.txt) | above + core earth science analysis packages | ![](https://img.shields.io/docker/image-size/pangeo/pangeo-notebook?sort=date) | ![](https://img.shields.io/docker/pulls/pangeo/pangeo-notebook?sort=date)
+| [pytorch-notebook](pytorch-notebook/packages.txt) | above + GPU-enabled pytorch | ![](https://img.shields.io/docker/image-size/pangeo/pytorch-notebook?sort=date) | ![](https://img.shields.io/docker/pulls/pangeo/pytorch-notebook?sort=date)
 | [ml-notebook](ml-notebook/packages.txt) | above + GPU-enabled tensorflow2 | ![](https://img.shields.io/docker/image-size/pangeo/ml-notebook?sort=date) | ![](https://img.shields.io/docker/pulls/pangeo/ml-notebook?sort=date)
```
weiji14, line 17: Instead of saying 'above', maybe better to explicitly say that `pytorch-notebook` extends `pangeo-notebook`? I got a bit confused by this...
```suggestion
| [pangeo-notebook](pangeo-notebook/packages.txt) | base-notebook + core earth science analysis packages | ![](https://img.shields.io/docker/image-size/pangeo/pangeo-notebook?sort=date) | ![](https://img.shields.io/docker/pulls/pangeo/pangeo-notebook?sort=date)
| [pytorch-notebook](pytorch-notebook/packages.txt) | pangeo-notebook + GPU-enabled pytorch | ![](https://img.shields.io/docker/image-size/pangeo/pytorch-notebook?sort=date) | ![](https://img.shields.io/docker/pulls/pangeo/pytorch-notebook?sort=date)
| [ml-notebook](ml-notebook/packages.txt) | pangeo-notebook + GPU-enabled tensorflow2 | ![](https://img.shields.io/docker/image-size/pangeo/ml-notebook?sort=date) | ![](https://img.shields.io/docker/pulls/pangeo/ml-notebook?sort=date)
```

Alternatively, we can have a diagram similar to https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#image-relationships
image

pytorch-notebook/environment.yml
scottyhq: > clusters don't pull the latest pangeo-docker image? 

Actually they do currently pull the latest :) https://github.com/2i2c-org/infrastructure/search?q=ml-notebook 
But I say we go ahead and merge this, and can use explicit pins on the hub if need be going forward

README.md
scottyhq: "pytorch-notebook extends pangeo-notebook " sounds good.  A diagram would be great, but let's do that later:  If directly in the repository readme can use this https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid

pytorch-notebook/environment.yml
weiji14: Wow, you all like to live dangerously :laughing: So looking at https://github.com/2i2c-org/infrastructure/pull/1244, it seems like the GPUs are K80s, and according to https://forums.developer.nvidia.com/t/nvidia-tesla-k80-cuda-version-support/67676/6, CUDA 11 to work but support is deprecated (i.e. there might be lots of warnings). So let's test it out then!

scottyhq (approved):

pytorch-notebook/environment.yml
rabernat: We are not currently using ml-notebook in producing in 2i2c. All production clusters use pinned tags.

README.md
weiji14: Thanks for the link to that! I hadn't realize GitHub had added this already, previously I was using https://yuml.me but it's much nicer to have an in-built solution. Anyways, I've added the diagram in another commit, looks like this:

image

pangeo-notebook/environment.yml
ngam: fyi, the conda-forge pytorch will pull in the appropriate* blas, currently set mkl, see more here: https://github.com/conda-forge/pytorch-cpu-feedstock/blob/93333de760d702aee276dcafeedc890458a00417/recipe/meta.yaml#L79-L84

So this will likely not be useful and get overridden anyway. Conda-forge also has a matrix of blas options, so I would just get it from there.

* if you go upstream (pytorch), you will see in their own default builds, they often have precedence for mkl (mkl being more performative than others, though unfortunately not exactly open and doesn't work on all platforms). I would drop this line

tests/test_pytorch-notebook.py
```diff
@@ -0,0 +1,26 @@
...
+
+packages = [
+    # machine learning stuff
+    "pytorch_lightning",
+    "skimage",
+    "sklearn",
+    "torch",
+    "torchgeo",
+    # cupy import fails unless on GPU-enabled node:
+    #'cupy', #libcuda.so.1: cannot open shared object file: No such file or directory
```
rabernat, line 14:
```suggestion
```

Let's remove this comment. It's just residue from old stuff in ml-image.

rabernat (commented):
This is great @weiji14 - just one small nit, then LGTM.

pangeo-notebook/environment.yml
rabernat: I could not get conda-lock to work without this explicitly in the environment (see https://github.com/pangeo-data/pangeo-docker-images/issues/312#issue-1215961431). If we can remove it, great--let's try! But that was not consistent with my experience. See also https://github.com/NVIDIA/MinkowskiEngine/issues/354.

pytorch-notebook/environment.yml
scottyhq: One other point on cudatoolkit 10 vs 11 is that 11 brings along a few more gigabytes. I don't mind big images, but this one is currently approaching 10GB total uncompressed...

`mamba create -n test --dry-run pytorch-gpu cudatoolkit=10` -> Total download: 1 GB
`mamba create -n test --dry-run pytorch-gpu cudatoolkit=11` ->  Total download: 3 GB

pangeo-notebook/environment.yml
ngam: if it works, it works! :)

tests/test_pytorch-notebook.py
ngam: Could add a suggestion as well? Because of the mkl and nomkl conflicts, this is sort of worrying... 

Try a test for the underlying blas to make sure things are set up correctly:

Usually it should be MKL (it's pulling MKL from conda-forge):

```python
import torch
assert torch.__config__.show()[torch.__config__.show().find('BLAS_INFO')+10:torch.__config__.show().find('BLAS_INFO')+13] == "mkl"
```

pangeo-notebook/environment.yml
ngam: I suggested an additional test below just in case

tests/test_pytorch-notebook.py
weiji14: Ok, done both in 3acef32a4f614726a502068446bd6c93e891dfb9. Let's see if the CI passes.

pytorch-notebook/environment.yml
weiji14: Yes cudatoolkit is huge unfortunately. There are ways to trim down the docker image size as you mentioned in https://github.com/pangeo-data/pangeo-docker-images/issues/188#issuecomment-788447708, but it's tough getting around a big binary... I think CUDA 11 will need to be used eventually though, so will need to figure out a solution :upside_down_face:

tests/test_pytorch-notebook.py
weiji14: Yep, works like a charm: `tests/test_pytorch-notebook.py::test_torch_uses_mkl PASSED               [100%]` https://github.com/pangeo-data/pangeo-docker-images/runs/6202609085?check_suite_focus=true#step:7:41

scottyhq:
Thanks @weiji14 for putting this together and @ngam @rabernat for the reviews, going to merge so that it'll be easier for people to test drive

Issue #302, opened 2022-04-15
ML notebook versions tightly pinned


scottyhq:
As of image 2022.04.15:

https://github.com/pangeo-data/pangeo-docker-images/blob/793b4d01621e1f3177991313044059f7506481cf/ml-notebook/environment.yml#L7-L10 

Discussion in the PR https://github.com/pangeo-data/pangeo-docker-images/pull/300 , but opening an issue to note this in case somebody wants to figure out how to increase compatible versions of jax, jaxlib, and tensorflow

ngam:
This should be generally be fixed on the conda-forge side for recent versions. The issue from the discussion seems to be the convoluted versioning the jax team follows (to facilitate better development, they iterate the simpler jax a lot more than they iterate the heavier and more complicated jaxlib, so the versioning can be quite hard to understand, but credit to them, they have thorough documentation on this: https://jax.readthedocs.io/en/latest/design_notes/jax_versioning.html)

PR #325, opened 2022-05-12
Add jax notebook

rabernat:
This is not working for me locally, but it's a start.

cc @dhruvbalwada


scottyhq:
Similar to the discussion with the new pytorch image: Do people use jax and tensorflow together, or is it typically one or the other? In an ideal world these libraries (tensorflow, pytorch, jax, cupy) could all have compatible dependencies, and not be giant, so we could have a single environment... but that does not seem to be the case currently (#302). In any case, if adding a separate jax notebook, do we remove jax from the tensoflow image:

https://github.com/pangeo-data/pangeo-docker-images/blob/e56a5fea09fb52362f53ed0f82735673c97b88e3/ml-notebook/environment.yml#L8-L10

ngam:
https://anaconda.org/ngam/jaxlib/files

I uploaded jaxlib with cuda for py37, py38, p39, and py310 here. I am fairly certain the actual product is okay, but we may need to fine tune the build process a little more.

`conda create -n jax39 python=3.9 jaxlib==*=*cuda* jax -c ngam -c conda-forge`. Note this is only cuda112 (11.2+).

ngam:
> Similar to the discussion with the new pytorch image: Do people use jax and tensorflow together, or is it typically one or the other?

I would advocate for coupling jax and tensorflow. Jax, in many ways, is an offshoot of tensorflow (if you look at the build logs, they're largely similar; jax borrows from tensorflow)

ngam:
> In an ideal world these libraries (tensorflow, pytorch, jax, cupy) could all have compatible dependencies, and not be giant, so we could have a single environment... but that does not seem to be the case currently (#302). In any case, if adding a separate jax notebook, do we remove jax from the tensoflow image?:

fwiw, you should be able to get that sorted. The problem here isn't in these libraries per se, it is in the many other deps we have. Here's how I would approach it: Start with these four and then add others to see where things break. For example, while torch and tensorflow could easily coexist, adding torchgeo breaks that --- why? Because in conda-forge, tensorflow is now iwth icu 70.1 but torchgeo is with gdal which is stuck at icu <70.1. But that will at some point get fixed. 

When I was working on the other containers based on NGC, I notice problems start popping up with random packages that I am not sure if they are useful to anyone --- or rather, I have never heard about them, so I have no idea what to make of them. A question is then, if some small package breaking down all compatibility, is it worth having? Who's using it and why? Could we replace it with something else? Unfortunately, I cannot make these calls because I am relatively new to the whole ecosystem, but I am sure people here will be able to figure it out.

ngam:
I wanna add, fixing these discrepancies and conflicts is often a simple PR away or a simple nudge to the feedstock maintainers, so it is not like reinventing the whole thing, just a bit of nudging...

ngam:
gdal icu70 PR: https://github.com/conda-forge/gdal-feedstock/pull/632 so hopefully things should be smoother soon-ish

ngam:
chased the gdal issue... now rasterio... after that xlayers... it's an endless game with these geo pkgs 😰

rabernat:
We can definitely drop xlayers! But not rasterio.

PR #336, opened 2022-05-27
add separate jax notebook with cuda

ngam:


ngam:
Something is failing in the pytorch notebook even though it wasn't edited...

PR #345, opened 2022-06-13
cuda112 & add pytorch to ml-notebook & add cupy

### LINKED ISSUE #320: Use cudatoolkit=11 in both tensorflow and pytorch images
Currently tensorflow (ml-notebook) uses cudatoolkit=10, and the new pytorch image uses cudatoolkit=11. It would be good to keep those versions in sync. Full discussion here:
https://github.com/pangeo-data/pangeo-docker-images/pull/315#pullrequestreview-955549154

### LINKED ISSUE #322: Add cupy to ml notebooks
It should be easy to support cupy in both ml notebooks, no?

ngam:
fixes #320
fixes #322 

@rabernat @scottyhq this ~minimal~ PR takes tensorflow to cuda112, adds cupy, and adds pytorch in ml-notebook

Ideally, we should add `jaxlib==*=*cuda*` (~and maybe even `pytorch==*=*cuda*` if you want to just have one "ML" notebook~ added a proposal!) but that will have to wait a little longer

Helpful notes for review:
- `pytorch-gpu` is just a meta package and so don't worry about its disappearance.
- openblas need not be present in the case of pytorch (the conda-forge pytorch only uses openblas --- in a bad way actually, long story --- for the case of m1 macs)


rabernat:
Thank you so much for working on this! 🙏 

Can you help me understand--if we are adding pytorch to ml-notebook, is there still any point to maintaining the separate pytorch-notebook?

ngam:
> if we are adding pytorch to ml-notebook, is there still any point to maintaining the separate pytorch-notebook?

No, there is no point. I simply didn't want to propose that big change here because I wasn't sure if it would be a welcome idea (but we can definitely take do it!).

More generally, I don't see a need to worry about separating pytorch and tensorflow going forward. We will likely face against some issues from time to time, but the tensorflow and pytorch feedstocks are currently maintained by the same set of volunteers, so the migrations happens around the same time. In fact, the trickiest package here is `torchgeo` for now (because it has dependencies on things like gdal). 

In summary, if we can add a recent enough tensorflow on top of pangeo-notebook with all these geo packages, we should be able to add pytorch as well. As far as I am concerned, I would treat jax and tensorflow very similarly. I don't have that deep of an experience with the rapids-ai collection, but I can look into that next. My impression is that they're on the easier side.

I am still investigating what's blocking us from getting to tensorflow 2.8.1, but I think it is some deep conflict caused by icu and/or absl stuff. Once we get tensorflow 2.8.1, we will have higher performance than the NGC containers (in my testing) and we get the cuda version of jaxlib as well (they have similar migrations).

ngam:
Could @weiji14 weigh on the pytorch discussion? Do you have any issues with merging the two notebooks?

ngam:
Update: I would abandon this PR if people are happy to wait a few days for #346 --- that will bring everything to the very cutting edge. We are just waiting on rasterio now.

weiji14:
> Could @weiji14 weigh on the pytorch discussion? Do you have any issues with merging the two notebooks?

Nope, as long as pytorch is available somewhere! I do worry though about the larger docker file size (especially if the RAPIDS AI stuff is meant to be added later), but other than that, ok with simplifying maintenance down to just one ml-notebook image.

> In fact, the trickiest package here is `torchgeo` for now (because it has dependencies on things like gdal).

Yeah, for better or worse, the dev version of `torchgeo` is also pinning upper versions of lots of packages (see https://github.com/microsoft/torchgeo/pull/557/files) which might make things more difficult going forward :slightly_smiling_face:

ngam:
Image sizes:

```
pangeo/ml-notebook        PR          5b99cf0c4873   Less than a second ago   10.2GB
pangeo/pytorch-notebook   PR          71b06612a1d4   1 second ago             8.84GB
```

Will likely grow by a little bit in #346 (up to 10.5GB, I would guess)

PR #348, opened 2022-06-14
parallel condalock

### LINKED ISSUE #347: proposal to further parallelize workflows + cache docker builds
- we could have each conda-lock operation run in parallel: https://github.com/pangeo-data/pangeo-docker-images/blob/master/.github/workflows/CondaLock.yml
- we could use docker caching to make the tests faster (when things aren't changing)

ngam:
fixes #347


ngam:
Ah! Just noticed this comment in the workflow:
```
      # Could run as single step in parallel, but would complicate logs...

```

closing.

PR #353, opened 2022-07-13
Bump base image to newer Ubuntu

### LINKED ISSUE #352: Bump to ubuntu 22.04 after 22.04.1 is out - sometime after August 4
When to update to the next version of Ubuntu? That has been considered for jupyter/docker-stacks. I suggest we do the same for these images.

> Ubuntu 22.04.1 [is scheduled](https://wiki.ubuntu.com/Releases) to be released August 4, 2022. When the docker image representing 22.04.1 is available on dockerhub, I suggest we upgrade. Practically, I see this meaning that we wait to August 4 and then look for an update [here](https://github.com/docker-library/official-images/blob/8b39fbf55a83e736786daeca86379a83d8f417b0/library/ubuntu#L44-L47) and then [on dockerhub](https://hub.docker.com/_/ubuntu?tab=tags&page=1&name=22.04). If our tests then pass here I think its time to go for the upgrade!
> 
> To do this aligns with the practice Ubuntu suggest _for users of the previous Ubuntu LTS_: to upgrade when their first patch release has been made, so that seems like a reasonable practice also for us.

Quote from myself in https://github.com/jupyter/docker-stacks/pull/1689#issuecomment-1183154138.

yuvipanda:
It's time! I don't actually think we need to wait for 20.4.1 - I think that
is valid for desktop systems and VMs, but not for container images. 
Particularly, we don't use the subsysems that are the primary
reason for 'wait for first point release' - the kernel, filesystems and the desktop
environment.

Fixes #352

### DIFF
--- a/base-image/Dockerfile
+++ b/base-image/Dockerfile
@@ -1,5 +1,5 @@
 # Dockerfile for base image of all pangeo images
-FROM ubuntu:20.04
+FROM ubuntu:22.04
 # build file for pangeo images
 
 LABEL org.opencontainers.image.source=https://github.com/pangeo-data/pangeo-docker-images


scottyhq (approved):
thanks!

PR #354, opened 2022-07-16
Add groff, dependency for awscli

### LINKED ISSUE #216: Installing aws cli version 2 and/or its dependency for the `help` subcommand
Using an image built `FROM pangeo/pangeo-notebook:2021.05.15` I'm using the `aws` CLI.

1. I wonder if you think we should install version 2 of the CLI (curl, unzip, :scream:, move on filesystem, done, :nauseated_face:).
2. I also wonder if you think we should install `groff` (an apt package). Apparently it is a dependency to run `help` sub-commands of like `aws s3 help` but `groff` isn't available by default in the `pangeo-notebook` image.

### Example output with `aws` CLI v1 trying to use `help` sub-command

```
$ aws
Note: AWS CLI version 2, the latest major version of the AWS CLI, is now stable and recommended for general use. For more information, see the AWS CLI version 2 installation instructions at: https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html

usage: aws [options] <command> <subcommand> [<subcommand> ...] [parameters]
To see help text, you can run:

  aws help
  aws <command> help
  aws <command> <subcommand> help
aws: error: the following arguments are required: command



$ aws help

Could not find executable named "groff"
```

consideRatio:
`groff` is a declared dependency for the `awscli` as seen in
https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html.

Without `groff`, running `aws help` will lead to a failure.

Closes #216.

### Important review

Please review that it makes sense to add groff to all these `apt.txt` files (`.binder/`, `pangeo-notebook/`, `ml-notebook/`, `pytorch-notebook/`). I'm both confused about finding a `.binder` folder and its relevance, as well as considering that ml/pytorch somehow depends on the pangeo-notebook.

### DIFF
--- a/ml-notebook/apt.txt
+++ b/ml-notebook/apt.txt
@@ -1,2 +1,3 @@
 git
+groff
 vim
--- a/pangeo-notebook/apt.txt
+++ b/pangeo-notebook/apt.txt
@@ -1,3 +1,4 @@
 git
-vim
+groff
 tree
+vim
--- a/pytorch-notebook/apt.txt
+++ b/pytorch-notebook/apt.txt
@@ -1,2 +1,3 @@
 git
+groff
 vim


yuvipanda:
The `.binder` is actually a symlink to `pangeo-notebook/` so if you launch this repo on binder, it gives you `pangeo-binder/`. 

Inheritance in this repo is done *purely* just for conda packages, and handled via conda-lock (see  https://github.com/pangeo-data/pangeo-docker-images/blob/5ca372d4b4655fb604ba97f9c6632d4d984fc04c/Makefile#L30). Hence the need to repeat apt.txt files elsewhere.

PR #355, opened 2022-07-18
Add a pangeo-forge runner image

yuvipanda:
Adds a `pangeo/forge` image containing environment for pangeo-forge runs,
inherited off pangeo-notebook with the following additions:

1. `apache-beam[gcp]` installed to allow for [Apache Beam](https://beam.apache.org/)
    execution, as pangeo-forge is moving towards this.
2. Puts the `boot` program that apache-beam needs into the appropriate place
3. A custom `start` script that launches either the apache-beam boot program or acts 
    as a no-op based on weather apache-beam is calling the container or some other
    program (such as binder)
4. Installs the `pangeo-forge-recipes` package into forge, rather than pangeo-notebook,
    so users can test it in binder as the 'playground' where needed.

I've also added test cases to make sure these features continue to work. In addition,
the test harnass wasn't actually testing the `Dockerfiles` in the other images - it was
writing a custom `Dockerfile` and testing *that*. This only tested the `ONBUILD` 
triggers from the base image, but not any customizations. This has been fixed.,

I'm installing `apache-beam[gcp]` via `pip` rather than `apache-beam-with-gcp` from
`conda-forge` as the latter seems to be broken - with it, beam was complaining about
not being able to find the libraries needed to talk to Google Cloud Store. This probably
needs to be fixed in [conda feedstock](https://github.com/conda-forge/apache-beam-feedstock)
somehow.


yuvipanda:
Hold on, just testing and working with Charles at SciPy sprints

cisaacstern:
To test this image with a real Pangeo Forge recipe, we did the following:

1. Built this image locally (which is a self-contained version of the image proposed in this PR)

    ```Dockerfile
    FROM golang:1.18-bullseye

    ENV VERSION=2.40.0
    RUN wget https://github.com/apache/beam/archive/refs/tags/v${VERSION}.tar.gz -O beam.tgz
    RUN tar xf beam.tgz
    
    RUN cd beam-${VERSION}/sdks/python/container && \
        CGO_ENABLED=0 go build -o /beam-boot
    
    FROM pangeo/pangeo-notebook:2022.07.13
    
    # FIXME: Pin version
    RUN pip install apache-beam[gcp]==2.40.0 --no-cache
    
    COPY --from=0 /beam-boot /usr/local/bin/boot
    
    ENTRYPOINT ["/usr/local/bin/boot"]
    ```
2. Appended this code (adapted from [this example](https://github.com/apache/beam/blob/master/sdks/python/apache_beam/examples/wordcount.py)) to a recipe which we know works on Dataflow

    ```python
    # recipe.py
    # The following recipe is copy-and-pasted above this code:
    # https://github.com/pangeo-forge/gpcp-feedstock/blob/4794d968c7d455bd6eef857864a5821a56bf4817/feedstock/recipe.py

    import argparse
    import logging
    
    import apache_beam as beam
    from apache_beam.options.pipeline_options import PipelineOptions
    from apache_beam.options.pipeline_options import SetupOptions
    
    from fsspec.implementations.local import LocalFileSystem
    from pangeo_forge_recipes.recipes import setup_logging
    from pangeo_forge_recipes.storage import FSSpecTarget
    
    setup_logging()
    local_fs = LocalFileSystem()
    recipe.storage_config.target = FSSpecTarget(local_fs, "test.zarr")
    recipe.cache_inputs = False
    chained_transform = recipe.copy_pruned().to_beam()
    
    parser = argparse.ArgumentParser()
    _, pipeline_args = parser.parse_known_args()
    pipeline_options = PipelineOptions(pipeline_args)
    pipeline_options.view_as(SetupOptions).save_main_session = True
    
    with beam.Pipeline(options=pipeline_options) as p:
        p | chained_transform
    ```

3. Run this `recipe.py`, with the following command [adapted from here](https://beam.apache.org/documentation/runtime/environments/#running-pipelines) (where `IMAGE_URL` is the image built in step 1 above)

    ```console
    export IMAGE="my-repo/beam_python_sdk_custom"
    export TAG="X.Y.Z"
    export IMAGE_URL = "${IMAGE}:${TAG}"
    
    python recipe.py \
    --runner=PortableRunner \
    --job_endpoint=embed \
    --environment_type="DOCKER" \
    --environment_config="${IMAGE_URL}"
    ```

4. As [described here](https://cloud.google.com/dataflow/docs/guides/using-custom-containers#basic_testing_with_portablerunner), the resulting zarr store will be on disk _inside the stopped container_ used for this run. To extract it, we call

    ```console
    docker cp $STOPPED_CONTAINER_ID:test.zarr test.zarr
    ```

5. `test.zarr` is now copied to our local filesystem, and we can open it with xarray

cisaacstern:
TODOs:

- [ ] Work with @scottyhq to get this PR merged
- [ ] Get Scott's insight re: how we can bump the `pangeo-forge-recipes` for this `forge` image to generate new `forge` tags without disrupting the release cycle for other images in this repo
- [ ] Once released, use this image for the Pangeo Forge Sandbox here: https://github.com/pangeo-forge/sandbox/blob/binder/binder/Dockerfile
- [ ] Build on @ncclementi's suggestion to create a meta-package like https://github.com/coiled/coiled-runtime, (`pangeo-forge-runtime`?) which is a single conda-installable package that will bring in all of the same dependencies as this image, so that users have a conda-installable pathway for replicating the cloud worker environment.
- [ ] This meta-package could be used in the CI environments here https://github.com/pangeo-forge/pangeo-forge-recipes/tree/master/ci ?
- [ ] Upstream Yuvi's approach in this image to the Beam docs here: https://beam.apache.org/documentation/runtime/environments/#modify-existing-base-image

TheNeuralBit:
:wave: hi @yuvipanda - I'm an Apache Beam maintainer. @cisaacstern brought this to my attention at the Beam Summit. I wanted to ask - is there something Beam could do that would make this easier? We definitely want it to be easy to build custom containers that include scientific libraries like this.

yuvipanda:
Hello @TheNeuralBit! Excited to have you pop over here :)

What would really help is if *all* that is needed for an image to be runnable on beam is for it to have the apache-beam python package installed. I am one of the maintainers of [jupyterhub](https://github.com/jupyterhub/jupyterhub) where we support arbitrary images run on infrastructure, and the requirement is that the `jupyterhub` package is installed in the image and the `jupyterhub-singleuser` is a well-known command that is available in $PATH. This makes image building very flexible.

So in an ideal world, just installing `apache-beam` will provide the functionality that `boot` does now. This can be accomplished in two ways:

1. Ship the `boot` binary in the python package! This is actually quite doable and not a bad idea at all, and you can add this as part of your `setup.py`. Then you just need to ask for `apache-beam-boot` or similar to be in $PATH.
2. Rewrite the `boot` binary in python.

Either of these would be very very helpful!

scottyhq:
> What would really help is if all that is needed for an image to be runnable on beam is for it to have the apache-beam python package installed 

If I understand the basic motivation for this PR it is to circumvent this limitation right? If so, I'd suggest opening an issue in this repository to revisit simplifying the `forge` configuration in the future

.github/workflows/Test.yml
```diff
@@ -53,13 +53,11 @@ jobs:
...
       run: |
         docker load < ./artifact/base-image.tar.gz
-        # Create temporary Dockerfile
-        echo "FROM ${DOCKER_ORG}/base-image:PR" > Dockerfile
 
     - name: Build Image
       run: |
         cd ${{ matrix.IMAGE }}
-        docker build -t ${DOCKER_ORG}/${{ matrix.IMAGE }}:PR -f ../Dockerfile .
+        docker build --build-arg  PANGEO_BASE_IMAGE_TAG=PR -t ${DOCKER_ORG}/${{ matrix.IMAGE }}:PR .
```
scottyhq, line 60: oh nice, this is much better than the temporary dockerfile hack

yuvipanda:
> If I understand the basic motivation for this PR it is to circumvent this limitation right? If so, I'd suggest opening an issue in this repository to revisit simplifying the `forge` configuration in the future

Pretty much. @TheNeuralBit do you think this would be welcome? Would love help in navigating that community and its dynamics!

yuvipanda:
I don't want to block this PR on that though.

forge/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+# 2. A 'beam boot' executable binary set as the entrypoint. This is a go program that
+#    actually launches the appropriate python code.
+#
+# These two need to be the same version to avoid any possible incompatibilites.
+#
+# In this image, we install the python package via pip (as the conda version seems currently
+# broken), and build the go executable by downloading the source & using a builder image
+# to build it. This ensures we meet the requirements needed for running beam while still
+# reusing most of the pre-existing infra we have for maintaining images.
+FROM golang:1.18-bullseye
```
scottyhq, line 15:
```suggestion
FROM golang:1.18-bullseye as beamboot
```

cisaacstern:
I've linked back to here again in https://github.com/apache/beam/issues/22349#issuecomment-1190654944, just so these two threads don't end up diverging. @TheNeuralBit, thanks so much for your engagement on this!

forge/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+FROM pangeo/base-image:${PANGEO_BASE_IMAGE_TAG}
+
+# This version is specified twice in this Dockerfile, so if you change this one,
+# make sure to change the other one too!
+ARG BEAM_VERSION=2.40.0
+
+# Install the same version of the package that we install the boot binary of
+RUN pip install --no-cache apache-beam[gcp]==${BEAM_VERSION}
+
+COPY --from=0 /beam-boot /usr/local/bin/boot
```
scottyhq, line 39:
```suggestion
COPY --from=beamboot /beam-boot /usr/local/bin/boot
```

i think this is clearer, feel free to use a different name though

forge/apt.txt
```diff
@@ -0,0 +1 @@
+git
```
scottyhq, line 1: maybe we should just install `git` from conda-forge in all the images...

yuvipanda: @scottyhq i'd actually recommend against that unless we have a specific reason, due to https://github.com/conda-forge/git-feedstock/issues/113 :(

pangeo-notebook/environment.yml
```diff
@@ -59,7 +58,6 @@ dependencies:
  - numpy
  - pandas
  - panel
- - pangeo-forge-recipes
```
scottyhq, line 62: if removing these you can either run conda-lock locally to add a new lockfile, or just update all images with the `/condalock` in a comment in this PR.

forge/apt.txt
scottyhq: ah good to know!

scottyhq:
> how we can bump the pangeo-forge-recipes for this forge image to generate new forge tags without disrupting the release cycle for other images in this repo 

This would require some conditionals in the current CI scripts. On average it's been a new build ~once a week and a new tag every two weeks (for all images), so maybe we just stick with that for the time being and revisit this in the future.

yuvipanda:
@scottyhq that's for the tagged (with dates) images right? Images with just commit hashes can be as frequent as necessary right? And the bakery can just use that when needed.

scottyhq (approved):
Looks good to me @yuvipanda and @cisaacstern! Thanks for all the comments and tests!

scottyhq:
> that's for the tagged (with dates) images right? Images with just commit hashes can be as frequent as necessary right? 

Correct!

yuvipanda:
@scottyhq looks like the pytorch image might be failing with some of the newer dependencies :( I'm going to try and revert the lock update across images and just do it for the forge.

forge/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+FROM golang:1.18-bullseye as beamboot
+
+# This version is specified twice in this Dockerfile, so if you change this one,
+# make sure to change the other one too!
+ARG BEAM_VERSION=2.40.0
+
+RUN curl -sSL https://github.com/apache/beam/archive/refs/tags/v${BEAM_VERSION}.tar.gz | tar -xzf -
+
+RUN cd beam-${BEAM_VERSION}/sdks/python/container && \
+    CGO_ENABLED=0 go build -o /beam-boot
```
alxmrs, line 24: Why not pull this from the Beam Python SDK directly? Are you sure you need to compile the go sources again? I'm not sure this is necessary.

yuvipanda: @abkfenris yeah, we could pull the go binary alone out of the beam sdk image (assuming that is what you are referring to) and I think that should be ok too. My initial reading of the beam dockerfile assumed that the python package was also installed into `/opt/apache/beam` and I didn't want to pull *that* in. But I guess I can copy just the binary out?

Do you know if it's statically built or links to local libc?

cisaacstern: Knowing @abkfenris, he'll probably be interested in this conversation, but you probably meant to tag @alxmrs. 😄

abkfenris: You mean I don't have to go down this rabbit hole right now? I mean I want to, but I really can't.

forge/Dockerfile
```diff
@@ -41,4 +26,4 @@ ARG BEAM_VERSION=2.40.0
 # Install the same version of the package that we install the boot binary of
 RUN pip install --no-cache apache-beam[gcp]==${BEAM_VERSION}
 
-COPY --from=beamboot /beam-boot /usr/local/bin/boot
+COPY --from=apache/beam_python3.8_sdk:2.40.0 /opt/apache/beam/boot /opt/apache/beam/boot
```
cisaacstern, line 29: Does it matter if you pull from `python3.8_sdk` vs. `python3.9_sdk`?

yuvipanda: @cisaacstern i'll switch it to 3.9 but it shouldn't matter as I'm picking up just the built go binary. Should modify docs too.

forge/Dockerfile
yuvipanda: @cisaacstern do you know what error you were getting?

forge/Dockerfile
alxmrs: Yes, it does matter: the python versions need to be consistent all the way through, or else you get Dataflow runtime errors. There is an open JIRA ticket in the beam project to address this, but I'm having a hard time finding it now.

yuvipanda: @alxmrs I'm only copying the `boot` go binary, not the built python package. The python package is being installed with `pip` into our conda environment directly here. I don't want to bring in any python dependencies by copying out of the beam image.

yuvipanda:
Looks like pytorch's newest version is currently broken - https://github.com/pytorch/pytorch/issues/74473. I'm gonna temporarily pin it to unblock this PR

yuvipanda:
@alxmrs ok, I'm now pulling just the go binary from the upstream docker image! It's no longer statically built, but I think in our specific case it's probably ok because we're still within the debian family. But I'm always wary of libc incompatibilities, so maybe worth making the build be static upstream too. 

Thank you so much for engaging with us here! If there is anything else that you think we could do here, please let us know?

yuvipanda:
I opened https://github.com/pangeo-data/pangeo-docker-images/issues/358 to unpin pytorch when that's done

forge/conda-linux-64.lock
```diff
@@ -0,0 +1,635 @@
...
+https://conda.anaconda.org/conda-forge/noarch/fonts-conda-ecosystem-1-0.tar.bz2#fee5683a3f04bd15cbd8318b096a27ab
+https://conda.anaconda.org/conda-forge/linux-64/binutils_linux-64-2.36-hf3e587d_10.tar.bz2#9d5cdbfe24b182d4c749b86d500ac9d2
+https://conda.anaconda.org/conda-forge/linux-64/libgcc-ng-12.1.0-h8d9b700_16.tar.bz2#4f05bc9844f7c101e6e147dab3c88d5c
+https://conda.anaconda.org/conda-forge/linux-64/abseil-cpp-20210324.2-h9c3ff4c_0.tar.bz2#baa652d7d0da41d757a31d00b4ae2c38
+https://conda.anaconda.org/conda-forge/linux-64/aom-3.3.0-h27087fc_1.tar.bz2#fe863d1e92331e69c8f231df5eaf5e16
+https://conda.anaconda.org/conda-forge/linux-64/aws-c-common-0.6.2-h7f98852_0.tar.bz2#ce69a062b3080485b760378841240634
+https://conda.anaconda.org/conda-forge/linux-64/bzip2-1.0.8-h7f98852_4.tar.bz2#a1fd65c7ccbf10880423d82bca54eb54
+https://conda.anaconda.org/conda-forge/linux-64/c-ares-1.18.1-h7f98852_0.tar.bz2#f26ef8098fab1f719c91eb760d63381a
+https://conda.anaconda.org/conda-forge/linux-64/charls-2.3.4-h9c3ff4c_0.tar.bz2#c3f85a96a52befc5e41cab1145c8d3c2
+https://conda.anaconda.org/conda-forge/linux-64/cudatoolkit-10.2.89-h713d32c_10.tar.bz2#e6e944f6fcd9dfd7fe67128693f6b771
```
scottyhq, line 37: @yuvipanda i think this lock file is incorrect and actually corresponds to the ML notebook (cuda, tensorflow), i didn't see anything wrong with CI so maybe it got overwritten locally by accident

scottyhq:
```
cd forge
conda-lock lock --mamba -k explicit -f environment.yml -f ../pangeo-notebook/environment.yml -f ../base-notebook/environment.yml -p linux-64
```
doesnt have the ML libraries

yuvipanda: @scottyhq quite possible, i'm relocking!

forge/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+# Refer to the base-image/Dockerfile for documentation.
+FROM pangeo/base-image:${PANGEO_BASE_IMAGE_TAG}
+
+# This version should match what is specified in the COPY instruction. We can't use
+# ARG or ENV here unfortunately as those aren't valid in COPY instructions.
+RUN pip install --no-cache apache-beam[gcp]==2.40.0
+
+# Copy *just* the golang based boot binary from the beam sdk image, *not*
+# the entire python package they build there. We install the python bits ourselves.
+COPY --from=apache/beam_python3.9_sdk:2.40.0 /opt/apache/beam/boot /opt/apache/beam/boot
```
scottyhq, line 28: Could be a place to use `COPY --link` if we eventually switch to dockerfile1.4 syntax (#357) https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#copy---link

forge/conda-linux-64.lock
yuvipanda: @scottyhq looks like the lockbot isn't locking forge yet, might need this to merge first. I'll try do it locally.

scottyhq: Yeah the `/condalock` command runs like `pull-request-target` so it doesn't see the new action, you have to run just the forge lock locally and then after merging it will work through CI

forge/Dockerfile
cisaacstern: TIL `COPY --link`. The docker knowledge on this repo/team are awesome 💪 !

yuvipanda:
@scottyhq i've updated the lockfile (and made sure the packages.txt file contains pip components too), and merged it! Thanks a lot for review + approval!

PR #360, opened 2022-07-26
Fixes pangeo forge image to be functional in Apache Beam environment 

yuvipanda:
- Fiddle with the `start` script, so it launches the beam boot binary correctly. Had to
  fix how we were detecting the boot binary being called, as well as how we were calling
  `exec` - args' first param needs to be the name of the command for commandline parsing
  to work properl
- GCP DataFlow unfortunately expects the container to run as root - it mounts some directories
   under `/var/opt` as root and expects the launched process to be able to read / write to it :( This
   we should raise upstream and find some way to fix - but in the meantime, the forge image now
   defaults to root. This shouldn't affect Jupyter use cases though, as we set the uid explicitly and
   ignore what the image says most of the time.


scottyhq (approved):

yuvipanda:
Thanks @scottyhq

Issue #306, opened 2022-04-20
Upgrade to conda-lock>1.0 


scottyhq:
Conda-lock >=1.0 added the ability to handle both conda-forge and pip packages! (however, separate solvers are used and compatible packages not guaranteed so it's best to stick with conda-forge as much as possible)
https://github.com/conda-incubator/conda-lock#pip-support

It would be nice though to just have a single conda-lock.yml per image rather than separate `conda-linux64.lock` and `requirements.txt`. It would also enable solving for a macos environment in addition to just linux.

This would require updating the main Dockerfile to use `conda-lock install` instead of separate `mamba install` and `pip install` steps.

PR #362, opened 2022-07-27
Test that apache_beam is installed in the forge image

yuvipanda:
conda-lock doesn't actually install any pip packages I
believe.


yuvipanda:
Something super funky is going on here.

```
$ mamba install -c conda-forge apache-beam-with-gcp
$ mamba list | grep beam
apache-beam               2.40.0           py39hb9d737c_0    conda-forge
apache-beam-with-gcp      2.40.0           py39hf3d152e_0    conda-forge
$ python3
Python 3.9.13 | packaged by conda-forge | (main, May 27 2022, 16:56:21) 
[GCC 10.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import apache_beam
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'apache_beam'
>>> 
```

yuvipanda:
As expected, https://github.com/pangeo-data/pangeo-docker-images/runs/7547348518?check_suite_focus=true fails - https://github.com/conda-forge/apache-beam-feedstock/pull/52 should fix it!

yuvipanda:
https://github.com/pangeo-data/pangeo-docker-images/pull/362/commits/2542609d25955c3f371485da71a29a77246fa610 made it pass, yay! So I think https://github.com/conda-forge/apache-beam-feedstock/pull/52 will definitely fix it I think.

scottyhq:
> conda-lock doesn't actually install any pip packages I believe.

In theory after v1.0 it does https://conda-incubator.github.io/conda-lock/pip/ ! But support for all the pip syntax isn't fully supported. This appears to work though: 

*Use new `conda-lock.yml` format to deal with both conda and pip packages:*
```
conda-lock lock -p linux-64
conda-lock install -n test
# Note also you can render conda-compatible lockfiles `conda-lock render -k env`, but there is a warning about how these don't necessarily preserve identical environments
```

I think it could be worth supporting `conda-lock.yml` files here:
https://github.com/pangeo-data/pangeo-docker-images/blob/3d0d6eeae9e6ded756bc5f6137aace43151524c8/base-image/Dockerfile#L156
see also #306  

But also it's great to fix the conda-forge version not mix with pip, so thanks for doing that!

scottyhq (approved):

yuvipanda:
Thanks a lot, @scottyhq! I removed the explicit dependency as my conda-forge PR was merged, and the tests pass now! \o/

yuvipanda:
Agree that switching to the yaml format is probably better!

PR #365, opened 2022-08-11
unpin PyTorch

### LINKED ISSUE #358: Unpin pytorch dependency
pytorch latest version seems to be broken as of a few dayas ago: https://github.com/pytorch/pytorch/issues/74473

I've pinned it in https://github.com/pangeo-data/pangeo-docker-images/pull/355 so that PR can proceed. We should unpin it when the issue is resolved.

ngam:
Fixes #358


pytorch-notebook/packages.txt
```diff
@@ -520,7 +520,7 @@ seaborn==0.11.2
 seaborn-base==0.11.2
 segmentation-models-pytorch==0.2.1
 send2trash==1.8.0
-setuptools==63.4.2
+setuptools==59.5.0
```
scottyhq, line 523: this seems like a pretty big drop, any idea why?

PR #378, opened 2022-09-30
add cuda-nvcc to tensorflow image + update versions to latest jax/tf

### LINKED ISSUE #377: Jax with GPU support and Flax or Equinox
Wondering where the debate on the having Jax is for now? Maybe it makes sense to have a separate notebook with jax, and no tensorflow? I am not an expert, would be curious to know how to use jax with gpu's on these images. 

I saw that Jax is part of the ml-notebook, but does not seem to be working with gpus. 
Screen Shot 2022-09-27 at 12 31 54 PM

Also, with jax notebooks it would be good to have support for some of the associated neural network libraries, like [flax](https://flax.readthedocs.io/en/latest/index.html) or/and [equinox](https://docs.kidger.site/equinox/). 
At the moment I was not able to get flax to work by pip installing it, as it changed the version of jax (I think it needs atleast [0.3.2](https://github.com/google/flax/issues/2135)).

ngam:
Fix #377 

Also add cuda-nvcc to get ptxas to be readily available in the environment (xla relies on ptxas to work correctly). Note missing ptxas would not have necessarily resulted in any errors in tf/jax, but without ptxas (cuda-nvcc), xla simply doesn't work, so it would go back to unoptimized performance.

ptxas isn't available in conda-forge at the moment. It is available in some images (i.e. any image built on top of NVIDIA's NGC images) and it is available on their conda channel (nvidia). I am told it is a licensing issue, but I don't really know. There will likely be a move to bundle it inside jax going forward, but that's beyond me.


dhruvbalwada (approved):
Thank you so much for doing this. The new image seems to be working as I would expect.

scottyhq (approved):
> Also add cuda-nvcc to get ptxas to be readily available in the environment (xla relies on ptxas to work correctly) 

As someone not familiar with these libraries I have no idea what cuda-nvcc, ptxas, and xla are :) But @ngam thanks for doing this and feel free to add any tests you'd like in a follow up!

PR #389, opened 2022-10-12
pin cuda-nvcc as a temporary workaround

### LINKED ISSUE #387: Using jax in ml-notebook instantly kills kernel
I am trying to use jax on the ml-notebook, but run into the problem that even the most basic functions kill the kernel. 

If I run the following code (first 4 lines from the [jax tutorial](https://jax.readthedocs.io/en/latest/notebooks/quickstart.html)): 
```
import jax.numpy as jnp
from jax import grad, jit, vmap
from jax import random

key = random.PRNGKey(0)
```

I get the error that kernel is restarting. I am using the m2lines deployment.

Since, I don't even know what the error is, I am not sure who is the person to reach out to about this. I will tag @jmunroe from 2i2c, as his team is most well-versed in this hub.

ngam:
fix #387 (potentially, pending more testing)


scottyhq (approved):

yuvipanda:
@ngam @dhruvbalwada is this *only* to support k80? We can just say we no longer support k80s and switch just to T4s. In https://github.com/2i2c-org/infrastructure/pull/1772 I make T4s the default

dhruvbalwada:
This PR will provide support for T4s only. The problem with K80s was not resolved and requires turning multi-threading off, so maybe just switch over to T4 entirely.

yuvipanda:
@dhruvbalwada yeah, https://github.com/2i2c-org/infrastructure/pull/1772 makes T4s the default - K80s are still available there. We only have quota for 4 T4s, should we get more?

dhruvbalwada:
I don't know the answer to that one yet. I personally probably won't be using more than that (probably will be using only 1 GPU) for the next few months. Maybe if we need more, then we can increase at a later point.

Also, based on looking at the logs, do you see any other users reaching the limit often? If so, maybe we can ask them personally if they would like more?

rabernat:
> Edit: I actually didn't even realize that more than one GPU can be used. How do we access them?

These quotas are for the whole project. It means only 4 people can be using T4s at once on the _entire M2LInES hub_. We don't support multiple GPUs per user, but could potentially try that if people thought it would be useful.

In terms of general usage, we have a bit of chicken vs. egg problem here. We have not done much work to explain to users how to take advantage of these resources. We don't have demos, examples, training, etc. So it is not surprising that usage is low.

In the short term, it's fine to drop K80s and use only T4s. I doubt that we would have more than 4 simultaneous GPU users in the project.

yuvipanda:
We could definitely try multiple GPUs on the same pod if folks think that would be useful. I've made T4s the default now, but I think the suggestion is that we drop K80s completely. https://cloud.google.com/compute/docs/gpus#nvidia_gpus_for_compute_workloads is the list of GPUs available, let me know if we should make any other ones available. It is fairly trivial to do so!

Issue #390, opened 2022-10-14
Document ML-image tag/GPU type/CUDA compatibility table


scottyhq:
> I think it would be very useful to have a table of cuda versions and the pangeo docker images they are compatible with. 

from @yuvipanda in https://github.com/pangeo-data/pangeo-docker-images/issues/387#issuecomment-1276732810 

The images have cuda libraries that are only compatible with certain GPUs (K80, T4, etc)
https://github.com/pangeo-data/pangeo-docker-images/blob/118d49782bac9486d07bcc1b6231067473bcfa7b/ml-notebook/packages.txt#L96-L97

Would be great to document this in the [readme](https://github.com/pangeo-data/pangeo-docker-images#dask-gateway-compatibility) @ngam or @weiji14 any chance you'd like to create a short compatibility table? Or add a short message and link to the relevant NVIDIA docs?

weiji14:
The K80s (Kepler generation, 2014) are a special case since later generation NVIDIA CUDA drivers (Maxwell onwards) have better forward compatibility with cudatoolkit, see https://docs.nvidia.com/deploy/cuda-compatibility/#forward-compatibility-title. Given the discussion in https://github.com/2i2c-org/infrastructure/issues/1765 on updating all the Pangeo hub clusters to T4s (can't say I didn't see this coming), I think we can probably just leave it at that (unless we expect to be supporting K80s for much longer).

Pro tip: The general rule of thumb [I follow](https://github.com/weiji14/deepbedmap/commit/4dd92c8f29b007619ca905a338ffb7050f5157ec) is to always have the latest CUDA **driver** version, which should work with any cuda**toolkit** (CTK) version 6.5 or above (though I recommend at least cudatoolkit 9.0).

![CUDA compatibility](https://user-images.githubusercontent.com/23487320/195747231-dd7f7dbe-54e0-4439-94f7-1ae1eeedde24.png)

yuvipanda:
I'd say in that case the suggestion would be to document *preferred GPUs*, and say folks should prefer T4 (if that is actually the case - I don't know!). I'm not super well versed in GPU usage, so would love some guidance that says 'just use these GPUs' from folks in the knoew.

weiji14:
> @weiji14 do you have access to these hubs? If not, the best person to document this is going to be @dhruvbalwada who will have hands-on experience with the systems/hubs in question.

If you mean the m2lines hubs, then no. Probably best to ask @dhruvbalwada.

[...]

PR #393, opened 2022-10-17
Update README.md

dhruvbalwada:
document the pinning of cuda-nvcc. Hoping to close: https://github.com/pangeo-data/pangeo-docker-images/issues/390

@ngam - could you refine this note more, if needed? Or tell me what else might be helpful to add here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,7 @@ The runtime environment sets two variables by default
 
 ### Other notes
 
+* Around 2022.10.14, the cuda-nvcc was pinned to 10.6.124, to match the latest cuda version (11.6) that could be installed on the m2lines hubs. This was done to make sure that JAX would work without throwing an error (https://github.com/pangeo-data/pangeo-docker-images/issues/387). 
 * Since 2020.10.16, [mamba](https://github.com/mamba-org/mamba) is installed into the base-image and conda-lock environment and is used by default to solve for a compatible environment (see #146)
 * For a simple list of packages for a given image, you can use a link like this: https://github.com/pangeo-data/pangeo-docker-images/blob/2020.10.08/pangeo-notebook/packages.txt
 * To compare changes between two images, you can use a link like this: https://github.com/pangeo-data/pangeo-docker-images/compare/2020.10.03..2020.10.08


README.md
```diff
@@ -167,6 +167,7 @@ The runtime environment sets two variables by default
 
 ### Other notes
 
+* Around 2022.10.14, the cuda-nvcc was pinned to 10.6.124, to match the latest cuda version (11.6) that could be installed on the m2lines hubs. This was done to make sure that JAX would work without throwing an error (https://github.com/pangeo-data/pangeo-docker-images/issues/387).
```
scottyhq, line 170: 10.6.124 -> 11.6.124 ?

scottyhq:
thanks @dhruvbalwada! What about putting this comment in a new section at the bottom of the readme like "GPU compatibility" 

I think it's good to convey "we recommend using T4 or newer NVIDIA GPUs in order to be compatible with the cudatoolkit>=11 versions installed in pangeo ML images"

README.md
weiji14:
```suggestion
* Around 2022.10.14, the cuda-nvcc was pinned to 10.6.124, to match cudatoolkit version 11.6 that could be installed on the m2lines hubs. This was done to make sure that JAX would work without throwing an error (https://github.com/pangeo-data/pangeo-docker-images/issues/387). 
```

Probably be specific that 11.6 refers to the cudatoolkit version (yes, the terminology around cuda drivers, cudatoolkit and cuda-nvcc is confusing...).

Do we want to mention something about the minimum CUDA driver version required (e.g. `450.36.06`) even though it's a bit irrelevant for these docker images?

yuvipanda: I don't think we should make a specific reference to the m2lines hub. Instead, may I suggest something about *avoiding* K80s, as they are a bit old and need version matching per https://github.com/pangeo-data/pangeo-docker-images/issues/390#issuecomment-1278395143? I also am not sure if putting cudatoolkit (or cuda-nvcc?) to latest and testing on T4 would help. Maybe the forward compatibilty measured in https://github.com/pangeo-data/pangeo-docker-images/issues/390#issuecomment-1278395143 means it would be fine? At which point we can just say 'K80 not supported' and be done.

README.md
```diff
@@ -167,6 +167,7 @@ The runtime environment sets two variables by default
 
 ### Other notes
 
+* Around 2022.10.14, the cuda-nvcc was pinned to 11.6.124, to match cudatoolkit version 11.6. This was done to make sure that JAX would work without throwing an error (https://github.com/pangeo-data/pangeo-docker-images/issues/387).
```
weiji14, line 170: Actually, I just double checked and [`cuda-nvcc-11.6.124`](https://github.com/pangeo-data/pangeo-docker-images/blob/770b758a579444f511ab259ed04df7ed77c061ee/ml-notebook/conda-linux-64.lock#L7) is matched with [`cudatoolkit=11.7`](https://github.com/pangeo-data/pangeo-docker-images/blob/770b758a579444f511ab259ed04df7ed77c061ee/ml-notebook/conda-linux-64.lock#L37) in the lockfile. I'm guessing the `nvidia-smi` output at https://github.com/pangeo-data/pangeo-docker-images/issues/387#issuecomment-1276804420 which indicates `CUDA Version: 11.6` is the cuda(toolkit?) version installed on Kubernetes, and `cudatoolkit=11.7` is the one from the docker image?

To be honest, all of these minor version differences (11.6, 11.7) shouldn't matter according to https://stackoverflow.com/questions/53422407/different-cuda-versions-shown-by-nvcc-and-nvidia-smi, since cuda 11 has gotten better with forward and backward compatibility compared to cuda 10. But this sentence probably needs a bit more rewording still :upside_down_face:

dhruvbalwada: @weiji14 - I am very new and confused about the terminology here. Could you take a stab at rewording? 
@ngam - had described the main problem here: https://github.com/pangeo-data/pangeo-docker-images/issues/390#issuecomment-1278996477. We had to downgrade cuda-nvcc so it would match the driver version on the hub. I don't know what sets the highest driver number that is avaiable, as @yuvipanda was the one who sets this somewhere (?)?

weiji14: Can you first report the cuda **driver** version on the M2Lines Hub using `nvidia-smi`? I see at https://github.com/pangeo-data/pangeo-docker-images/issues/387#issuecomment-1276804420 that @yuvipanda reported it as `Driver Version: 510.47.03`, but want to confirm if that is still the case. This driver (the connection between the NVIDIA GPU hardware and all the CUDA software libraries) basically dictates what versions of stuff can be used.

dhruvbalwada: Ok, will do. 
Is this not set by ?: 
[cuda-nvcc==11.6.124
cudatoolkit==11.7.0](https://github.com/pangeo-data/pangeo-docker-images/blob/118d49782bac9486d07bcc1b6231067473bcfa7b/ml-notebook/packages.txt#L96-L97)

dhruvbalwada: At the moment it reads: 
Screen Shot 2022-10-18 at 11 22 52 AM

weiji14: Thanks, if you have access to the K80 GPUs, can you also paste the output of `nvidia-smi` from there?

dhruvbalwada: @ngam - I am leaning on the same side as you about this now. 
If this is only a problem for Jax, then maybe we can have the default be an unpinned version. Then leave a note that that if someone wants to solve the jax issue then they can mamba install the version number after checking driver number. What do you all think?

dhruvbalwada:
This has been solved in https://github.com/pangeo-data/pangeo-docker-images/pull/398.

PR #398, opened 2022-10-18
remove cuda-nvcc and document ptxas

### LINKED ISSUE #390: Document ML-image tag/GPU type/CUDA compatibility table

ngam:
Closes #393 
Fixes #390

cc @weiji14 @yuvipanda @dhruvbalwada @scottyhq


dhruvbalwada:
Should it be removed (used can install if needed) or just unpinned (user downgrades if needed)? Just wondering if unpinning is better, as some other things might be using it?

README.md
```diff
@@ -170,6 +170,7 @@ The runtime environment sets two variables by default
...
 * To compare changes between two images, you can use a link like this: https://github.com/pangeo-data/pangeo-docker-images/compare/2020.10.03..2020.10.08
+* Our `ml-notebook` image now contains JAX and TensorFlow with XLA enabled. Due to licensing issues, conda-forge does not have `ptxas`, but `ptxas` is needed for XLA to work correctly. Should you like to use JAX and/or TensorFlow with XLA optimization, please install `ptxas` on your own, for example, by `conda install -c nvidia cuda-nvcc`. At the time of writing (October 2022), JAX throws a compilation error if the `ptxas` version is higher than the driver version. There does not exist an easy solution for K80 GPUs, but in the case of T4 GPUs, you should install `conda install -c nvidia cuda-nvcc==11.6.*` to be safe. Alternatively for any GPU, you could set an environment variable to resolve the error caused by JAX: `XLA_FLAGS="--xla_gpu_force_compilation_parallelism=1"`. The aforementioned error will be removed (and likely turned into a warning) in a future version of JAX.
```
weiji14, line 173: Cross-referencing https://github.com/google/jax/issues/12776#issuecomment-1276649134. Seems like there are some similar issues mentioned e.g. in https://github.com/google/jax/issues/11967

```suggestion
* Our `ml-notebook` image now contains JAX and TensorFlow with XLA enabled. Due to licensing issues, conda-forge does not have `ptxas`, but `ptxas` is needed for XLA to work correctly. Should you like to use JAX and/or TensorFlow with XLA optimization, please install `ptxas` on your own, for example, by `conda install -c nvidia cuda-nvcc`. At the time of writing (October 2022), JAX throws a compilation error if the `ptxas` version is higher than the driver version. There does not exist an easy solution for K80 GPUs, but in the case of T4 GPUs, you should install `conda install -c nvidia cuda-nvcc==11.6.*` to be safe. Alternatively for any GPU, you could set an environment variable to resolve the error caused by JAX: `XLA_FLAGS="--xla_gpu_force_compilation_parallelism=1"`. The aforementioned error will be removed (and likely turned into a warning) in a future version of JAX. See https://github.com/google/jax/issues/12776#issuecomment-1276649134
```

yuvipanda:
From a cluster perspective, I'm happy to keep drivers up to date with whatever it is the cloud providers support (which is out of our hands). If we feel this isn't going to affect non-jax users, and they have a well documented workaround, I think this seems fine to me. 

Based on all this conversation, 2i2c now uses T4s as default GPUs for everyone https://github.com/2i2c-org/infrastructure/pull/1787. Thanks everyone!

weiji14 (approved):

Issue #396, opened 2022-10-18
Images availability for OS/ARCH other than linux/amd64


j34ni:
@scottyhq 

I would like to use your docker images (in particular pangeo/pangeo-notebook and pangeo/ml-notebook) on an IBM POWER9 machine and I was wondering if there was any plan to make these available for linux/ppc64le?

scottyhq:
> I was wondering if there was any plan to make these available for linux/ppc64le 

Unfortunately, there is no plan for that. We only support and test `linux/amd64`, mainly because that is the main cloud architecture available for the Pangeo-supported JupyterHubs these images are designed for.

That said, I'd be interested to hear how it goes if anybody does put some time into building these images for other platforms! https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

j34ni:
@scottyhq 

That is a shame, because beside this POWER9 machine I was planing to run these images on ARM-based machines (and I guess that others may also be interested)

I started to rebuild with `buildx build --platform=linux/ppc64le` and that seems to work for the `base-image`
The only changes I had to make to the Dockerfile were `FROM ppc64le/ubuntu:22.04` and `Mambaforge-Linux-ppc64le.sh`

However the `pangeo-notebook` failed at 
``` 
=> ERROR [7/1] RUN echo "Checking for pip 'requirements.txt'..."                     
```
Any idea about how I should adapt that?

yuvipanda:
I do think we should provide ARM builds though :D

scottyhq:
There are two parts for this, and I suggest starting without docker and just getting a conda environment for your architecture:

1) Do the conda packages exist (either cross-platform or platform-specific)
2) If yes, build a docker image that encapsulates that conda environment

I quickly tried re-creating the pangeo-notebook environment on AWS (`c6g.medium` -> `amazon/ubuntu/images/hvm-ssd/ubuntu-jammy-22.04-arm64-server-20220912`) with no luck because of various packages that I suspect have compiled dependencies:

```
wget https://github.com/conda-forge/miniforge/releases/latest/download/Mambaforge-Linux-aarch64.sh  
bash Mambaforge-Linux-aarch64.sh  
mamba env create --file https://raw.githubusercontent.com/pangeo-data/pangeo-docker-images/master/pangeo-notebook/environment.yml
```

```
Encountered problems while solving:
  - nothing provides requested ciso
  - nothing provides requested esmpy
[...]

PR #399, opened 2022-10-19
Add support for multi-arch docker images

yuvipanda:
- Install appropriate version of mambaforge based on build architecture
- Generate lockfiles on `base-notebook` and `pangeo-notebook` for:
  - osx-64 (x86_64)
  - osx-arm64 (Apple Sillicon M1/M2/M3 chips)
  - linux-aarch64 (arm64)
  - ~ppc64le (POWER8/9)~

Ref https://github.com/pangeo-data/pangeo-docker-images/issues/396


yuvipanda:
conda-lock also needs to know though.

.github/workflows/CondaLock.yml
```diff
@@ -50,51 +50,66 @@ jobs:
            activate-environment: condalock
            use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
 
+      - name: Specify which architectures to build lockfiles for
+        run: |
+          # space delimited list of architectures to send to conda-lock
+          echo "BUILD_ARCHS='64 aarch64 ppc64le'" > ${GITHUB_ENV}
```
scottyhq, line 56: i think it'd be best to just start with one additional? aarch64 and add others down the line if need be

.github/workflows/CondaLock.yml
```diff
@@ -50,51 +50,66 @@ jobs:
...
       - name: Run conda-lock base-notebook
         timeout-minutes: 5
         continue-on-error: true
         run: |
           conda clean --all --yes --force-pkgs-dirs
           cd base-notebook
-          conda-lock lock --mamba -k explicit -f environment.yml -p linux-64
-          ../generate-packages-list.py conda-linux-64.lock > packages.txt
+          for ARCH in ${BUILD_ARCHS}; do
+            conda-lock lock --mamba -k explicit -f environment.yml -p linux-${ARCH}
```
scottyhq, line 66: this might be a good time to switch to a unified lockfile and actually add conda-lock to the `base` environment. (Our current usage of conda-lock pre-dates the unified lockfile format now shown here https://github.com/conda-incubator/conda-lock#basic-usage). Installing the environment with conda-lock would actually recognize the host platform and choose the correct set of packages:

```
conda-lock lock -f environment.yml -p linux-64 aarch64
conda-lock install -n pangeo-notebook
```

the drawback is that if someone just has conda or mamba installed they dont recognize the new unified lockfile format... so you could not do `conda env create -f conda-lock.yml` ...

yuvipanda: @scottyhq ok that's pretty cool and we clearly should just move to that.

Makefile
```diff
@@ -12,38 +12,58 @@ base-image :
 base-notebook : base-image
 	cd base-notebook ; \
 	conda-lock lock --mamba -k explicit -f environment.yml -p linux-64; \
-	../generate-packages-list.py conda-linux-64.lock > packages.txt; \
+	../generate-packages-list.py conda-linux-64.lock > packages-64.txt; \
+	conda-lock lock --mamba -k explicit -f environment.yml -p linux-aarch64; \
+	../generate-packages-list.py conda-linux-aarch64.lock > packages-aarch64.txt; \
+	conda-lock lock --mamba -k explicit -f environment.yml -p linux-ppc64le; \
+	../generate-packages-list.py conda-linux-ppc64le.lock > packages-ppc64le.txt; \
```
scottyhq, line 19: see comment above on unified lockfile. but i still really like a simple text file as well though! maybe we create a `artifacts` subdirectory to organize files not actually used by docker build?

Makefile
```diff
@@ -12,38 +12,58 @@ base-image :
...
 .PHONY: forge
 forge : base-image
 	cd forge; \
 	conda-lock lock --mamba -k explicit -f environment.yml -f ../pangeo-notebook/environment.yml -f ../base-notebook/environment.yml -p linux-64; \
-	../generate-packages-list.py conda-linux-64.lock > packages.txt; \
+	../generate-packages-list.py conda-linux-64.lock > packages-64.txt; \
+	conda-lock lock --mamba -k explicit -f environment.yml -f ../pangeo-notebook/environment.yml -f ../base-notebook/environment.yml -p linux-aarch64; \
+	../generate-packages-list.py conda-linux-aarch64.lock > packages-aarch64.txt; \
+	conda-lock lock --mamba -k explicit -f environment.yml -f ../pangeo-notebook/environment.yml -f ../base-notebook/environment.yml -p linux-ppc64le; \
+	../generate-packages-list.py conda-linux-ppc64le.lock > packages-ppc64le.txt; \
```
scottyhq, line 67: TBH I have rarely used this Makefile and am always just building things via GitHub Actions ;) I've put in a little more time to embracing `docker buildx` and think it might be the right tool to evolve this repo. in particular the `bake` tool for several related images https://github.com/scottyhq/pangeo-buildx#usage

yuvipanda: @scottyhq I completely agree - I love *that* repo and think that's what this should be :D What do you think needs to happen to complete that transition?

scottyhq: Awesome! Mainly just more time :) I abandoned it for a bit with the "dont fix what isn't broken" rationale :) but it's also nice to keep evolving so I'll try to open a PR here this week and would love your review

.github/workflows/CondaLock.yml
weiji14: Let's build just for `linux-aarch64` first as mentioned at https://github.com/pangeo-data/pangeo-docker-images/issues/396#issuecomment-1643006035

.github/workflows/CondaLock.yml
weiji14: Thinking that we might need to update from conda-lock=1.4 to conda-lock=2.x first at https://github.com/pangeo-data/pangeo-docker-images/blob/554675c5269c1b8d634004f6a5529c4031fdeeb4/environment-condalock.yml#L6

Also FYI, `mamba` doesn't support installing from the unified `conda-lock.yml`, but `micromamba` does https://mamba.readthedocs.io/en/latest/user_guide/micromamba.html#conda-lock-yaml-spec-files. See also https://github.com/mamba-org/mamba/issues/1884#issuecomment-1245170833

scottyhq:
@yuvipanda @weiji14 I'm thinking we should go ahead and merge this. Likely there will be some things to work out, but linux and mac ARM support will be a nice addition

base-image/Dockerfile
```diff
@@ -66,8 +66,8 @@ USER ${NB_USER}
 WORKDIR ${HOME}
 
 # Install latest mambaforge in ${CONDA_DIR}
-RUN echo "Installing Miniforge..." \
-    && URL="https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-x86_64.sh" \
+RUN echo "Installing Mambaforge..." \
+    && URL="https://github.com/conda-forge/miniforge/releases/latest/download/Mambaforge-Linux-$(uname -m).sh" \
```
weiji14, line 70: Probably should keep using Miniforge instead of Mambaforge which is discouraged since Sep 2023 (https://github.com/conda-forge/miniforge/pull/488). Doesn't look like they're deprecating Mambaforge anytime soon, but don't want to get caught up in that.

pangeo-notebook/environment.yml
```diff
@@ -13,7 +13,7 @@ dependencies:
  - cdsapi
  - cfgrib
  - cf_xarray
- - ciso
+ - ciso
```
weiji14, line 16: Remove trailing blankspace here.
```suggestion
 - ciso
```

weiji14 (commented):
> @yuvipanda @weiji14 I'm thinking we should go ahead and merge this. Likely there will be some things to work out, but linux and mac ARM support will be a nice addition

Yep, agree with going ahead with this. Let me just test running `conda-lock` locally on some of these images to check that there are no serious issues. Also a few minor suggestions below :point_down:

.github/workflows/CondaLock.yml
```diff
@@ -37,39 +37,37 @@ jobs:
...
-
-      - name: Upload lockfiles
-        uses: actions/upload-artifact@v4
-        with:
-          name: ${{ matrix.image }}
-          path: ${{ matrix.image }}
+          cd base-notebook
+          conda-lock lock -p linux-64 -p linux-aarch64 -p osx-64 -p osx-arm64
+	        conda-lock render -k explicit -p linux-64
+	        ../generate-packages-list.py conda-linux-64.lock > packages.txt
```
weiji14, line 43: Remove tabs to fix leading indentation
```suggestion
          conda-lock lock -p linux-64 -p linux-aarch64 -p osx-64 -p osx-arm64
          conda-lock render -k explicit -p linux-64
          ../generate-packages-list.py conda-linux-64.lock > packages.txt
```

.github/workflows/CondaLock.yml
```diff
@@ -37,39 +37,37 @@ jobs:
...
-          token: ${{ secrets.PANGEOBOT_TOKEN }}
-          # These lines are critical, otherwise Pangeo-bot pushes changes directly to master from PRs!
-          repository: ${{ github.event.client_payload.pull_request.head.repo.full_name }}
-          ref: ${{ github.event.client_payload.pull_request.head.ref }}
+      - name: Run conda-lock ml-notebook
+        timeout-minutes: 5
+        continue-on-error: true
+        run: |
+          cd ml-notebook
+          conda-lock lock -f environment.yml -f ../pangeo-notebook/environment.yml -f ../base-notebook/environment.yml -p linux-64 -p linux-aarch64 -p osx-64 -p osx-arm64
```
weiji14, line 59: Tensorflow doesn't have a linux-aarch64 package yet - https://github.com/conda-forge/tensorflow-feedstock/issues/136

```
Could not lock the environment for platform linux-aarch64
Could not solve for environment specs
The following package could not be installed
└─ tensorflow >=2.15.0 cuda120* does not exist (perhaps a typo or a missing channel).
```

.github/workflows/CondaLock.yml
```diff
@@ -37,39 +37,37 @@ jobs:
...
+          ../generate-packages-list.py conda-linux-64.lock > packages.txt
 
-      # Download all artifacts from previous matrix job
-      - uses: actions/download-artifact@v4
+      - name: Run conda-lock pytorch-notebook
+        timeout-minutes: 5
+        continue-on-error: true
+        run: |
+          cd pytorch-notebook
+          conda-lock lock -f environment.yml -f ../pangeo-notebook/environment.yml -f ../base-notebook/environment.yml -p linux-64 -p linux-aarch64 -p osx-64 -p osx-arm64
```
weiji14, line 68: Pytorch does have linux-aarch64/osx-64/osx-arm64 packages on conda-forge, but we set a `cuda120` pin in the environment.yml file so locking fails:

```
Could not lock the environment for platform linux-aarch64
Could not solve for environment specs
The following packages are incompatible
├─ pytorch >=2.1.0 cuda120* is not installable because it requires
│  └─ __cuda, which is missing on the system;
└─ torchvision >=0.16.1 cuda120* does not exist (perhaps a typo or a missing channel).
```

weiji14 (commented):
Ok, tested things locally, and locking on `base-notebook` and `pangeo-notebook` worked, but `ml-notebook` and `pytorch-notebook` didn't. Maybe we just do `base-/pangeo-` for now in this PR, and I can do a follow-up one for `ml-/pytorch-`?

base-image/Dockerfile
scottyhq: good catch!

.github/workflows/Test.yml
```diff
@@ -42,6 +42,9 @@ jobs:
         sudo rm -rf /usr/local/lib/android /usr/share/dotnet /opt/ghc
         df -h
 
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v3
```
scottyhq, line 46: following https://docs.docker.com/build/ci/github-actions/multi-platform/.

scottyhq:
Added some commits to update the Makefile and Actions to actually build for multiple platforms, and it seems to be working!

base-image/Dockerfile
```diff
@@ -144,8 +144,10 @@ ONBUILD USER ${NB_USER}
 # We want to keep our images as reproducible as possible. If a lock
 # file with exact versions of all required packages is present, we use
 # it to install packages. conda-lock (https://github.com/conda-incubator/conda-lock)
-# is used to generate this conda-linux-64.lock file from a given environment.yml
-# file - so we get the exact same versions each time the image is built. This
+# is used to generate this conda-lock.yml file from a given environment.yml
+# file - so we get the exact same versions each time the image is built. Since this is
+# different packages for different CPU architectures
+# the same dockerfile can be used to build different architecture images. This
```
weiji14, line 150: Sentence read a little funny. I guess we're trying to say that while different CPU architectures will require packages that are compiled differently, we can still use the multi-platform conda-lock.yml file for the Docker build?

```suggestion
# file - so we get the exact same versions each time the image is built. Note that
# different packages may be used for different CPU architectures, but still,
# the same dockerfile can be used to build different architecture images. This
```

weiji14 (approved):
Pre-approving since this is mostly ready, but have a look at the suggestions first. We'll probably need to disable `osx-64` for `ml-notebook` and `pytorch-notebook` because of the (Linux-only) CUDA pin set in the environment.yml file.

PR #426, opened 2023-01-12
Don't delete *.pyc files from the image

yuvipanda:
Panel and Xarray-Leaflet are heavy enough imports that without .pyc files, they sometimes together take as much as 15s to import?! This causes jupyterhub to fail startup in some cases.

The longer term fix is in panel and xarray-leaflet ( see https://github.com/holoviz/panel/issues/4271,
https://github.com/xarray-contrib/xarray_leaflet/issues/79).

In the meantime, leaving the .pyc files in place doesn't increase the image size by much, but makes startup definitely much faster!

Ref https://github.com/2i2c-org/infrastructure/issues/2047

### DIFF
--- a/base-image/Dockerfile
+++ b/base-image/Dockerfile
@@ -74,7 +74,7 @@ RUN echo "Installing Mambaforge..." \
     # to try reduce image size - see https://jcristharif.com/conda-docker-tips.html
     # Although we explicitly do *not* delete .pyc files, as that seems to slow down startup
     # quite a bit unfortunately - see https://github.com/2i2c-org/infrastructure/issues/2047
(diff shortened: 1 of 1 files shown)


scottyhq (approved):
Amazing analysis in the linked issue @yuvipanda this seems like it was a tricky one to track down! I've always been a bit reluctant about removing files to save space in the docker image (why not just remove them from the package if they aren't necessary?) 

This illustrates there are two related performance optimizations to track 1. container size to reduce time pulling and storing images but which impacts 2. the container start-up time in jupyterhub!

yuvipanda:
Thanks @scottyhq! Yes, the performance of the pull is the trade-off here! I think .pyc are small enough though, so is probably fine - although let’s see how big the image gets when it is built and pushed!

PR #433, opened 2023-01-19
Bump xarray-leaflet

yuvipanda:
Brings in https://github.com/xarray-contrib/xarray_leaflet/issues/79, which speeds up container startup quite a bit!


pangeo-notebook/environment.yml
```diff
@@ -89,7 +89,7 @@ dependencies:
  - xarray
  - xarrayutils
  - xarray-datatree
- - xarray_leaflet
+ - xarray_leaflet>=0.2.3
```
scottyhq, line 92: i think this should be 0.2.2? In any case, once it's on conda-forge (see linked issue) we can add it!

scottyhq:
closed by #431

PR #439, opened 2023-02-08
add xclim package

thenaomig:
This PR proposes to add the [xclim](https://xclim.readthedocs.io/en/stable/) package to the pangeo-notebook image. 

Suggesting to add this because xclim contains more than one [requested](https://discourse.pangeo.io/t/climate-indices-toolbox/2048) feature: climate indices and bias-correction, understanding that this violates the [principle](https://pangeo.io/packages.html#best-practices-for-pangeo-projects) of "a clear and relatively narrow scope". (Additionally, an aspect of xclim that may be duplicative is the handling of units, which MetPy also does.)

That said, the [bias-correction sub-module](https://xclim.readthedocs.io/en/stable/sdba.html), in particular, does not seem to exist elsewhere for the particular ability to do trend-preserving, quantile-based bias-correction (e.g. [QDM](https://journals.ametsoc.org/view/journals/clim/28/17/jcli-d-14-00754.1.xml)), in python, and in a package that builds on xarray. Bias-correction of climate model outputs is a general problem that broad groups of users may encounter.


rabernat (approved):
Thanks for this! 🙌

PR #440, opened 2023-02-09
Update environment.yml

Timh37:
Request for adding https://github.com/maximenc/pycop (multivariate dependence modeling) to Pangeo Cloud


scottyhq:
It looks like this package is not on conda-forge @Timh37 , if possible we prefer packages coming from conda-forge to not mix the dependency solving:

```
Could not solve for environment specs
Encountered problems while solving:
  - nothing provides requested pycop
```

Issue #417, opened 2022-12-14
condalock.yml workflow broken with conda 2.11.1


scottyhq:
documented the upstream issue here https://github.com/conda-incubator/conda-lock/issues/292 

unfortunately also because we continue-on-error the workflow indicates a pass even though lockfiles are not added
https://github.com/pangeo-data/pangeo-docker-images/actions/runs/3644277462

fix for now is to pin conda<22.11

PR #456, opened 2023-05-13
Pin minimum conda to 23.3.0

weiji14:
The condalock KeyError issue due to conda 22.11.1 reported in https://github.com/pangeo-data/pangeo-docker-images/issues/417 should have been fixed in conda 23.3.0 according to https://github.com/conda/conda-lock/issues/292#issuecomment-1490079666.


scottyhq:
There seems to be some issue with solving the pangeo-forge image:

@yuvipanda @cisaacstern would it be okay to remove these pins? 
```
https://github.com/pangeo-data/pangeo-docker-images/blob/7d4e51e81c47685cc6012a6b100f7ce023e3ccc9/forge/environment.yml#L7-L12
```

yuvipanda:
@scottyhq   `apache-beam-with-gcp=2.43.0`  or the gcsfs one? The gcsfs related one seems broken still :(

scottyhq:
> @scottyhq  apache-beam-with-gcp=2.43.0 or the gcsfs one? The gcsfs related one seems broken still :( 

Both, but I guess there was some network issue with locking previously that is now fixed... or new versions got released....

scottyhq (approved):

PR #474, opened 2023-07-13
Add a howto on determining packages content for an image

jmunroe:


jmunroe:
This PR is coming from
- https://github.com/2i2c-org/infrastructure/issues/2793

as part of the RTD documentation aimed at users of Pangeo Docker Images. It describes
- How to find my current image tag
- How to finding list of packages in any given tag
- How to finding diff of packages between any two tags

@yuvipanda comments?

docs/howto/determining-packages.md
```diff
@@ -0,0 +1,172 @@
...
+
+`pangeo/pangeo-notebook:2023.06.07`
+
+and
+
+`pangeo/pangeo-notebook:2023.06.20`
+
+to determine which packages were changed, added, or removed?
+
+### Using the command line with Bash
```
yuvipanda, line 111: Instead of using bash or python, we should allow people to do this directly from the GitHub UI. https://docs.github.com/en/pull-requests/committing-changes-to-your-project/viewing-and-comparing-commits/comparing-commits#comparing-tags has info. 

You can do so by:

1. Finding the release for the tag, from https://github.com/pangeo-data/pangeo-docker-images/tags
2. Select the tag, so you get something like https://github.com/pangeo-data/pangeo-docker-images/tags
3. Use the 'compare' dropdown on the left to select another tag to compare it against
4. Look at the packages.txt of the appropriate image.

IMO this is easier and more useful than having to write code to figure this out.

docs/howto/determining-packages.md
```diff
@@ -0,0 +1,172 @@
...
+
+Using GitHub the web site, the list of select a particular tag from the branches/tag drop down menu.
+
+## How to finding list of packages in any given tag
+
+Each image has an `packages.txt` file that lists exactly which packages and versions installed in the image.
+
+### Using GitHub website
+
+Using GitHub the web [site](https://github.com/pangeo-data/pangeo-docker-images), select a particular tag from the branches/tag drop down menu.  Navigate to the image and select the `packages.txt` file.
```
yuvipanda, line 57: Link directly to https://github.com/pangeo-data/pangeo-docker-images/tags?

docs/howto/determining-packages.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Using the command line with Bash
```
yuvipanda, line 59: IMO, just focusing on using the GitHub UI is good enough.

docs/howto/determining-packages.md
```diff
@@ -0,0 +1,172 @@
...
+| [ml-notebook](ml-notebook/packages.txt) | pangeo-notebook + GPU-enabled tensorflow2 | 
+| [forge](ml-notebook/packages.txt) | pangeo-notebook + [Apache Beam](https://beam.apache.org/) support| 
+
+Each image is also assigned a tag of the form `YYYY.MM.DD` indicating when it was last updated.
+
+*Click on the image name in the table above for a current list of installed packages and versions for the most recent tag of each image*
+
+## How to find my current image tag
+
+The environment variable `JUPYTER_IMAGE_SPEC` will tell you what Pangeo Docker Image is currently in use.
```
yuvipanda, line 19: My suggestion here is to show people with how to access `JUPYTER_IMAGE_SPEC` (`echo $JUPYTER_IMAGE_SPEC` in a terminal or even just `!echo $JUPYTER_IMAGE_SPEC` in a notebook), and then *describe* the various parts there. For example, the value may actually be of three forms:

1. `docker.io/pangeo-data/pangeo-notebook:tag`
2. `pangeo-data/pangeo-notebook:tag`
3. `quay.io/pangeo-data/pangeo-notebook:tag`

I think describing the registry (`docker.io`, which is often omitted, to mean dockerhub or `quay.io`), the repository (`pangeo-data/pangeo-notebook`) and the `tag` is cleaner and more helpful to the target audience I think.

docs/howto/determining-packages.md
```diff
@@ -0,0 +1,172 @@
+# Determining Packages in Pangeo Docker Images
+
+For deployments of Pangeo Docker Images using JupyterHub or Binder it can be sometimes unclear to the user exactly which image and tag is being used. Here are the images available as Pangeo Docker Images:
+
+| Image           | Description                                   | 
+|-----------------|-----------------------------------------------|
+| [base-notebook](base-notebook/packages.txt) | minimally functional image for pangeo hubs |
```
yuvipanda, line 7: Do these links work in the built sphinx documentation?

yuvipanda (changes requested):
Thanks, @jmunroe! I've left some comments, mostly about reducing the amount of code here and being a bit more descriptive. Thank you for working on this!

docs/howto/determining-packages.md
jmunroe: I've added instructions for using the GitHub UI

docs/howto/determining-packages.md
jmunroe: I think it depends on the user and what other tools they are used to using.

docs/howto/determining-packages.md
jmunroe: Good point. Updated to point to the corresponding GitHub url instead.

jmunroe:
Hi @yuvipanda . Thanks for the feedback on the changes. I've tried to address your comments.

jbusecke:
Thank you for working on this @jmunroe and @yuvipanda. Reading this I came across this statement:
> Given the LEAP hub exclusively uses [pangeo images](https://github.com/pangeo-data/pangeo-docker-images/)

Which is not strictly true currently (we offer the [education image](https://github.com/jbusecke/leap_education_docker_image) I made), and might become a lot less true now that we have the option to specify arbitrary images (which is **amazing**). I think this documentation here is great, but it is very specific to the pangeo-docker-images. Is there a more general way to list/diff *any* docker image? Or maybe there is some way to enable this to work at least for comparisons between pangeo images and ones created with the [2i2c template](https://docs.2i2c.org/admin/howto/environment/hub-user-image-template-guide/)?

To motivate this use-case a bit: I was planning on putting some time and energy into educating the LEAP community how to build their own images. I think that for the diverse needs of LEAP researchers it would often make sense to set up project-specific images maintained by single researchers/groups of researchers. I think upstreaming as much of this to 2i2c and other communities could be very nice.

yuvipanda (changes requested):

PR #476, opened 2023-08-02
added 3 packages to environment.yml

ds4114:
I work with the LDEO McKinley group and wanted to recommend 3 commonly used packages: cdsapi, cmocean, xgboost


pangeo-notebook/environment.yml
```diff
@@ -98,6 +100,7 @@ dependencies:
  - xclim
  # - xcube>=0.9.1
  - xesmf
+ - xgboost
```
weiji14, line 103: XGBoost is a machine learning algorithm, should this be added to the `ml-notebook` or `pytorch-notebook` instead? Are you using the CPU version of the GPU version of xgboost?

weiji14 (commented):
Hi @ds4114! Thanks for opening this Pull Request to add these oceanography related packages, just have one comment for now :smile:

pangeo-notebook/environment.yml
weiji14: Ah ok, if you're just using the CPU version, it should be fine to add xgboost to pangeo-notebook, similar to scikit-learn as you said. Let us know though if you decide to use the GPU version at some point (not sure if the LDEO JupyterHub has GPU clusters?), and we can add it to either `ml-notebook`/`pytorch-notebook` next time.

weiji14 (approved):
Looks ok to me, but @scottyhq, we'll need to see how to handle the unrelated no space left on device error on pytorch-notebook (similar to https://github.com/pangeo-data/pangeo-docker-images/pull/475#issuecomment-1646026710). Also, maybe we should check the docker image filesize if it's getting too big.

pangeo-notebook/environment.yml
weiji14: But `cdsapi` is on conda-forge no? See https://anaconda.org/conda-forge/cdsapi. Feel free to remove it though if you prefer to leave it out, and re-run `/condalock` afterwards.

weiji14: So we'll leave `cdsapi` in the docker image then?

rabernat: It makes sense to have cdsapi in the base notebook image. It is a general package for accessing ECMWF data.

PR #484, opened 2023-09-15
CI workflow maintenance

scottyhq:
Fix listing docker image size in CI logs, bump versions as well


.github/workflows/Binder.yml
scottyhq, line 1: in theory this workflow was pre-caching of pangeo-notebook on mybinder.org. But the tags were wrong and furthermore I don't think it's really necessary to pre-build (the first person just waits a bit longer)

weiji14 (approved):

PR #503, opened 2023-11-28
Add python-xxhash

dcherian:
This is what's needed for significantly faster hashing with dask, not just `xxhash`


jhamman (approved):

scottyhq:
@dcherian @jhamman are you able to provide a little more context for this? I see on this page https://docs.dask.org/en/stable/install.html?highlight=xxhash#optional-dependencies "Use xxHash hash functions for array hashing (~2x faster than MurmurHash, slightly slower than CityHash)" 

Seems pretty great if just having this in the environment speeds up any dask-array computation?

dcherian:
It'll speed up `dask.array.from_array(numpy_array)` type things, and potentially other hashing things but mostly that .

PR #514, opened 2024-02-16
Bump CUDA from 11.8 to 12.0

weiji14:
CUDA 12.0 migration across conda-forge is practically complete (see https://conda-forge.org/status/#cuda120), so we can start updating to a newer version of CUDA!

Note:
- CUDA 12.x requires CUDA driver >=525.60.13, see https://docs.nvidia.com/deploy/cuda-compatibility/index.html#cuda-intro and https://docs.nvidia.com/cuda/archive/12.0.0/cuda-toolkit-release-notes/index.html#cuda-toolkit-major-component-versions. Previous CUDA 11.x required CUDA driver >= 450.80.02.

Changes in this PR:
- [x] Update Pytorch, Torchvision and Tensorflow to use CUDA 12.0 builds
- [x] Update minimum pin on Tensorflow from >=2.14.0 to >=2.15.0
- [x] Update minimum pin on Pytorch from 2.0.0 to 2.1.0, and torchvision from 0.15.1 to 0.16.1
- [x] Pin minimum version of flax to 0.8.0, helps to resolve #489

References:
- https://developer.nvidia.com/blog/cuda-toolkit-12-0-released-for-general-availability
- https://github.com/conda-forge/conda-forge.github.io/issues/1963
- https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/4400

Supersedes #505, Fixes #489

PR #546, opened 2024-05-17
Bump conda-lock to 2.5, and increase locking timeout to 45min

weiji14:
Some updates cherry-picked from #514 to enable the CUDA upgrade from 11.8 to 12.0.

The 45min timeout might be a bit too generous, I can reduce it down to 30min or something after seeing what is actually required in #514 to lock the `pytorch-notebook` environment.

### DIFF
--- a/.github/workflows/CondaLock.yml
+++ b/.github/workflows/CondaLock.yml
@@ -14,7 +14,7 @@ jobs:
         IMAGE: [base-notebook, pangeo-notebook, ml-notebook, pytorch-notebook]
     name: ${{ matrix.IMAGE }}
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 45
     defaults:
       run:
         shell: bash -el {0}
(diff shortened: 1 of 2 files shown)


scottyhq:
> The 45min timeout might be a bit too generous, I can reduce it down to 30min or something after seeing what is actually 

Yeah, the default is 6 hours according to
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

Anything over a few minutes seems odd, I have no idea how these solvers work behind the scenes, but I wonder if it's worth opening an issue on conda-lock if it continues to spin. I had a quick look and only saw this possibly related one https://github.com/conda/conda-lock/issues/438

scottyhq (approved):

weiji14:
Seems like 45min didn't work either, so I'm increasing the timeout to 3 hours in #547. If that still doesn't work, I'll open an issue on https://github.com/conda-incubator/conda-lock

PR #556, opened 2024-06-24
Bump to python 3.12

### LINKED ISSUE #497: Upgrade Images to Python 3.12
**Is your feature request related to a problem? Please describe.**
https://docs.python.org/3/whatsnew/3.12.html

**Describe alternatives you've considered**
Leave everything at 3.11 ! 

**Additional context**
From past experience it takes packages a while to be compatible with the latest release of Python, so this is an issue to track

scottyhq:
closes #497


ml-notebook/packages.txt
```diff
@@ -45,8 +45,7 @@ aws-c-sdkutils==0.1.16
 aws-checksums==0.1.18
 aws-crt-cpp==0.26.12
 aws-sdk-cpp==1.11.329
-awscli==2.17.0
-awscrt==0.20.10
+awscli==1.32.106
```
scottyhq, line 48: https://github.com/conda-forge/awscli2-feedstock/pull/39

pangeo-notebook/environment.yml
```diff
@@ -112,7 +112,8 @@ dependencies:
  - xarray_leaflet
  - xarray-spatial
  - xbatcher
- - xcape
+ # https://github.com/conda-forge/xcape-feedstock/pull/22
+ #- xcape
```
scottyhq, line 116: dropping b/c blocking both #396 and #497 and doesn't seem very active. people can always install it on top of the default environment

ml-notebook/packages.txt
weiji14: Hmm, that python312 migration for awscli v2 doesn't look very active. Do we want to wait for that, or go with this awscli downgrade to v1?

scottyhq: i think downgrading is fine. i've still yet to see a situation that *requires* v2 (https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html)... As long as `aws s3 ls` works I'm happy :)

weiji14: Yeah, I don't do much fancy stuff either besides `aws s3 ls` and `aws s3 cp` :joy:. We'll hear soon enough if people have issues (and can let them pin to pangeo-docker-images `2024.06.23`:stuck_out_tongue_closed_eyes:)

weiji14 (approved):
Thanks @scottyhq!

PR #575, opened 2024-08-21
Colorize terminals even when home is overridden

### LINKED ISSUE #574: DRAFT ISSUE: Terminal colors etc -- adjust to missing /etc/skel files in user's home dir (`.profile`, `.bashrc`, `.bash_logout`)?
DRAFT ISSUE - I mean to think more about this, but have to return thinking about this

When linux users are created, they get files copied to their home directory copied from `/etc/skel`. These files aren't available in the image's default home directory, and even if they were, they would typically get replaced by a mounted home directory not bootstrapped with the files.

Due to this, we won't get colors in the terminals for example, and terminal headings are different.

Without /etc/skel files in home dir | With /etc/skel files in home dir
-|-
image|image

## Reference reading

- [Bash startup files](https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html)

## `/etc/skel` content reference

<details><summary><code>ls -alh /etc/skel/</code></summary>

```bash
root@ed803ea74248:/# ls -alh /etc/skel/ 
total 20K
drwxr-xr-x 2 root root 4.0K Aug  8 14:07 .
drwxr-xr-x 1 root root 4.0K Aug 21 09:03 ..
-rw-r--r-- 1 root root  220 Jan  6  2022 .bash_logout
-rw-r--r-- 1 root root 3.7K Jan  6  2022 .bashrc
-rw-r--r-- 1 root root  807 Jan  6  2022 .profile
```

</details>

<details><summary><code>cat .bash_logout</code></summary>

```bash
# ~/.bash_logout: executed by bash(1) when login shell exits.

# when leaving the console clear the screen to increase privacy

if [ "$SHLVL" = 1 ]; then
    [ -x /usr/bin/clear_console ] && /usr/bin/clear_console -q
fi
```

</details>

<details><summary><code>cat .bashrc</code></summary>

```bash
# ~/.bashrc: executed by bash(1) for non-login shells.
# see /usr/share/doc/bash/examples/startup-files (in the package bash-doc)
# for examples

# If not running interactively, don't do anything
case $- in
    *i*) ;;
      *) return;;
esac

# don't put duplicate lines or lines starting with space in the history.
# See bash(1) for more options
HISTCONTROL=ignoreboth

# append to the history file, don't overwrite it
shopt -s histappend

# for setting history length see HISTSIZE and HISTFILESIZE in bash(1)
HISTSIZE=1000
HISTFILESIZE=2000

# check the window size after each command and, if necessary,
# update the values of LINES and COLUMNS.
shopt -s checkwinsize

# If set, the pattern "**" used in a pathname expansion context will
# match all files and zero or more directories and subdirectories.
#shopt -s globstar

# make less more friendly for non-text input files, see lesspipe(1)
[ -x /usr/bin/lesspipe ] && eval "$(SHELL=/bin/sh lesspipe)"

# set variable identifying the chroot you work in (used in the prompt below)
if [ -z "${debian_chroot:-}" ] && [ -r /etc/debian_chroot ]; then
    debian_chroot=$(cat /etc/debian_chroot)
fi

# set a fancy prompt (non-color, unless we know we "want" color)
case "$TERM" in
    xterm-color|*-256color) color_prompt=yes;;
esac

# uncomment for a colored prompt, if the terminal has the capability; turned
# off by default to not distract the user: the focus in a terminal window
# should be on the output of commands, not on the prompt
#force_color_prompt=yes

if [ -n "$force_color_prompt" ]; then
    if [ -x /usr/bin/tput ] && tput setaf 1 >&/dev/null; then
	# We have color support; assume it's compliant with Ecma-48
	# (ISO/IEC-6429). (Lack of such support is extremely rare, and such
	# a case would tend to support setf rather than setaf.)
	color_prompt=yes
    else
	color_prompt=
    fi
fi

if [ "$color_prompt" = yes ]; then
    PS1='${debian_chroot:+($debian_chroot)}\[\033[01;32m\]\u@\h\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ '
else
    PS1='${debian_chroot:+($debian_chroot)}\u@\h:\w\$ '
fi
unset color_prompt force_color_prompt

# If this is an xterm set the title to user@host:dir
case "$TERM" in
xterm*|rxvt*)
    PS1="\[\e]0;${debian_chroot:+($debian_chroot)}\u@\h: \w\a\]$PS1"
    ;;
*)
    ;;
esac

# enable color support of ls and also add handy aliases
if [ -x /usr/bin/dircolors ]; then
    test -r ~/.dircolors && eval "$(dircolors -b ~/.dircolors)" || eval "$(dircolors -b)"
    alias ls='ls --color=auto'
    #alias dir='dir --color=auto'
    #alias vdir='vdir --color=auto'

    alias grep='grep --color=auto'
    alias fgrep='fgrep --color=auto'
    alias egrep='egrep --color=auto'
fi

# colored GCC warnings and errors
#export GCC_COLORS='error=01;31:warning=01;35:note=01;36:caret=01;32:locus=01:quote=01'

# some more ls aliases
alias ll='ls -alF'
alias la='ls -A'
alias l='ls -CF'

# Add an "alert" alias for long running commands.  Use like so:
#   sleep 10; alert
alias alert='notify-send --urgency=low -i "$([ $? = 0 ] && echo terminal || echo error)" "$(history|tail -n1|sed -e '\''s/^\s*[0-9]\+\s*//;s/[;&|]\s*alert$//'\'')"'

# Alias definitions.
# You may want to put all your additions into a separate file like
# ~/.bash_aliases, instead of adding them here directly.
# See /usr/share/doc/bash-doc/examples in the bash-doc package.

if [ -f ~/.bash_aliases ]; then
    . ~/.bash_aliases
fi

# enable programmable completion features (you don't need to enable
# this, if it's already enabled in /etc/bash.bashrc and /etc/profile
# sources /etc/bash.bashrc).
if ! shopt -oq posix; then
  if [ -f /usr/share/bash-completion/bash_completion ]; then
    . /usr/share/bash-completion/bash_completion
  elif [ -f /etc/bash_completion ]; then
    . /etc/bash_completion
  fi
fi
```

</details>

<details><summary><code>cat .profile</code></summary>

```bash
# ~/.profile: executed by the command interpreter for login shells.
# This file is not read by bash(1), if ~/.bash_profile or ~/.bash_login
# exists.
# see /usr/share/doc/bash/examples/startup-files for examples.
# the files are located in the bash-doc package.

# the default umask is set in /etc/profile; for setting the umask
# for ssh logins, install and configure the libpam-umask package.
#umask 022

# if running bash
if [ -n "$BASH_VERSION" ]; then
    # include .bashrc if it exists
    if [ -f "$HOME/.bashrc" ]; then
	. "$HOME/.bashrc"
    fi
fi

# set PATH so it includes user's private bin if it exists
if [ -d "$HOME/bin" ] ; then
    PATH="$HOME/bin:$PATH"
fi

# set PATH so it includes user's private bin if it exists
if [ -d "$HOME/.local/bin" ] ; then
    PATH="$HOME/.local/bin:$PATH"
fi
```

</details>

consideRatio:
Fixes #574.

The first commit accomplishes this when this image is used in for example JupyterHub deployments where the home folder is overridden and doesn't come with `/etc/skel` copied files `.profile` and `.bashrc`:

Before|After
-|-
image|image

This PR is also bundled with an opinionated change in another a second commit. It makes an opinionated adjustment of the terminal title and prompt. The title is set to the current working directory of the terminal, and the prompt is slimmed to not include the `@hostname` part.

Before | After
-|-
image|image

### DIFF
--- a/base-image/Dockerfile
+++ b/base-image/Dockerfile
@@ -80,6 +80,13 @@ RUN echo "Installing Miniforge..." \
     # quite a bit unfortunately - see https://github.com/2i2c-org/infrastructure/issues/2047
     && find ${CONDA_DIR} -follow -type f -name '*.a' -delete
 
+# Default to using /etc/skel/.profile and /etc/skel/.bashrc if not in the user's
+# home directory.
+USER root
+RUN echo ". /etc/bash.bashrc-skel-appendix" >> /etc/bash.bashrc
+COPY bash.bashrc-skel-appendix /etc/bash.bashrc-skel-appendix
+USER ${NB_USER}
+
 EXPOSE 8888
 ENTRYPOINT ["/srv/start"]
 #CMD ["jupyter", "notebook", "--ip", "0.0.0.0"]
--- a/base-image/bash.bashrc-skel-appendix
+++ b/base-image/bash.bashrc-skel-appendix
@@ -1,0 +1,27 @@
+# /etc/skel files should have been copied to a user's home directory, but if the
+# user's home directory is mounted for example it may not contain files initially
+# copied from /etc/skel. This can lead to terminals without colors etc.
+#
+# This script is made to be sourced from the end of /etc/bash.bashrc in order to
+# default to source files from /etc/skel as a fallback to missing files in the
+# home directory.
+#
+# ref: https://github.com/pangeo-data/pangeo-docker-images/issues/574.
+#
+
+# ~/.bash_profile, ~/.bash_login, or ~/.profile are defined, it doesn't make
+# sense to fallback to /etc/skel/.profile, so then return.
+[ -f "$HOME/.bash_profile" ] && return
+[ -f "$HOME/.bash_login" ] && return
+[ -f "$HOME/.profile" ] && return
+
+# /etc/skel/.profile directly sources ~/.bashrc if defined, so we first check if
+# it isn't, and then sources /etc/skel/.bashrc instead.
+#
+if [ ! -f "$HOME/.bashrc" ]; then
+    . /etc/skel/.bashrc
+fi
+
+
+# Now run /etc/skel/.profile as it wasn't in the user's home directory.
+. /etc/skel/.profile
--- a/base-image/bash.bashrc-skel-appendix
+++ b/base-image/bash.bashrc-skel-appendix
@@ -20,6 +20,49 @@
 #
 if [ ! -f "$HOME/.bashrc" ]; then
     . /etc/skel/.bashrc
+
+    # --------------------------------------------------------------------------
+    # The section below is adjusted from an Ubuntu 22.04 /etc/skel/.bashrc file
+    # in order to tweak how PS1 was set, influencing the terminal color, prompt
+    # and title.
+    #
+    # CHANGES:
+    # - removed "#" from the line with force_color_prompt=yes
+    # - removed "@\h" from PS1's prompt section
+    # - removed "\u@\h: " from the PS1's xterm title section
+    # --------------------------------------------------------------------------
+
+    # uncomment for a colored prompt, if the terminal has the capability; turned
+    # off by default to not distract the user: the focus in a terminal window
+    # should be on the output of commands, not on the prompt
+    force_color_prompt=yes
+
+    if [ -n "$force_color_prompt" ]; then
+        if [ -x /usr/bin/tput ] && tput setaf 1 >&/dev/null; then
+            # We have color support; assume it's compliant with Ecma-48
+            # (ISO/IEC-6429). (Lack of such support is extremely rare, and such
+            # a case would tend to support setf rather than setaf.)
+            color_prompt=yes
+        else
+            color_prompt=
+        fi
+    fi
+
+    if [ "$color_prompt" = yes ]; then
+        PS1='${debian_chroot:+($debian_chroot)}\[\033[01;32m\]\u\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ '
+    else
+        PS1='${debian_chroot:+($debian_chroot)}\u:\w\$ '
+    fi
+    unset color_prompt force_color_prompt
+
+    # If this is an xterm set the title to dir
+    case "$TERM" in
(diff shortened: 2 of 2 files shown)


consideRatio:
Some thoughts after making this.

1. Removing the `@hostname` from the prompt seemed reasonable, but I think it could be controversial - remove that change?
2. Trickery needs to be involved if the terminal title is to be retained as Terminal 1 etc, because its overridden by `/etc/skel` files, so I think stripping those from `username@hostname:` makes sense to include in this PR.
3. To apply or not apply the `/etc/skel.bashrc` stuff if a `.bashrc` file was found or not isn't obvious. I can imagine users are happy with colors, and then unhappy after defining a `.bashrc` file with some unrelated thing. Should it be applied no matter what?
4. Similarly as point 3, should we apply the /etc/skel stuff no matter what?
5. Should we apply /etc/skel files conditionally based on some environment variable, allowing this behavior to be opt-out or opt-in?

scottyhq:
Sorry to have missed this @consideRatio ! Agreed colors are nice and this is better than the default! I don't really have a strong preferences regarding your questions, but an alternative is maybe just pointing to 2i2c docs to customize the terminal (since hubs might use custom images not coming from here)? 

For what it's worth I've gotten quite used to copy and paste from stack overflow into ~/.bashrc whenever I log into a jupyterhub and don't find it too onerous :) 

```bash
# Colors and prompt customization
# https://unix.stackexchange.com/questions/148/colorizing-your-terminal-and-shell-environment
PS1='\e[34;1m\u@\h: \e[36m\W\e[0m\$ '

#export LS_COLORS='rs=0:di=01;34:ln=01;36:mh=00:pi=40;33'
LS_COLORS=$LS_COLORS:'di=1;35:' ; export LS_COLORS
export LS_OPTIONS='--color=auto'
alias ls='ls $LS_OPTIONS' 
```

consideRatio:
Closing as I'm out of capacity to follow through more on this :/

PR #579, opened 2024-09-11
Add exactextract

dcherian:
now on conda-forge! as of last week!

### DIFF
--- a/pangeo-notebook/environment.yml
+++ b/pangeo-notebook/environment.yml
@@ -26,6 +26,7 @@ dependencies:
  - eofs
  - erddapy
  - esmpy
+ - exactextract
  - fastjmd95
  - flox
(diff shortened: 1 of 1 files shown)


scottyhq:
Locking failing b/c there isn't an ARM build:
https://github.com/conda-forge/exactextract-feedstock?tab=readme-ov-file#current-release-info

dcherian:
sadly I just ran in to that too :/ is there a way to add it to platforms where possible?

scottyhq:
>  is there a way to add it to platforms where possible? 

@weiji14 went through this process for another package recently, documented below if someone wants to pursue it! 
https://github.com/pangeo-data/pangeo-docker-images/issues/396#issuecomment-2148329950

Issue #587, opened 2024-10-25
datatree now missing in 2024.10.11


dhruvbalwada:
**Describe the bug**
datatree is missing in the release: https://github.com/pangeo-data/pangeo-docker-images/releases/tag/2024.10.11 

It seems that datatree is not available in this release now. This may be partly related to the fact that datatree got merged into xarray (https://github.com/pydata/xarray/discussions/9680). 
However, this latest xarray release is not part of the current image. 

**Infrastructure (Where you are running this image):**
- leap.2i2c.cloud (@jbusecke is the owner)

jbusecke:
Thanks for raising this @dhruvbalwada. @scottyhq can we issue a new release here (which should include the new xarray)?

scottyhq:
>  @scottyhq can we issue a new release here (which should include the new xarray) 

The release process is just to add a new tag is just `git tag 2024.10.26` and `git push --tags`, which I just did, so hopefully it's fixed. @jbusecke in the future I think you have rights to do this?

jbusecke:
I seem to be able to issue a release from the gh interface, so maybe? Sorry I never checked for this.

PR #588, opened 2024-10-26
Install openssh so jupyter-sshd-proxy could work

yuvipanda:
Follow-up to https://github.com/pangeo-data/pangeo-docker-images/pull/585#event-14575680510.

The conda-forge package doesn't actually pull in openssh, which feels like the right call for now (as systems may already have openssh via other means?). 

Explicilty add openssh so jupyter-sshd-proxy works


yuvipanda:
Let me add a test before we merge this so we know for sure it is not broken.

pangeo-notebook/packages.txt
```diff
@@ -680,20 +687,20 @@ wayland-protocols==1.37
...
+webob==1.8.9
 websocket-client==1.8.0
 websockets==13.1
 wheel==0.44.0
 widgetsnbextension==4.0.13
 wrapt==1.16.0
 x264==1!164.3095
 x265==3.5
-xarray==2024.9.0
+xarray==2024.10.0
```
scottyhq, line 698: Also should close #587 !

scottyhq (approved):
thanks @yuvipanda !

PR #615, opened 2025-07-23
Add libgdal-arrow-parquet driver

weiji14:
Adding the `libgdal-arrow-parquet` driver for GDAL/OGR to support reading Geoparquet files.

Requested at https://github.com/NASA-IMPACT/pangeo-notebook-veda-image/issues/36

### DIFF
--- a/pangeo-notebook/environment.yml
+++ b/pangeo-notebook/environment.yml
@@ -65,6 +65,7 @@ dependencies:
  - jupyter-resource-usage
  - jupyter-sshd-proxy
  - kerchunk
+ - libgdal-arrow-parquet
  - line_profiler
  - lonboard
  - lxml


pangeo-notebook/environment.yml
```diff
@@ -65,6 +65,7 @@ dependencies:
  - jupyter-resource-usage
  - jupyter-sshd-proxy
  - kerchunk
+ - libgdal-arrow-parquet
```
weiji14, line 68: @wildintellect @jsignell, does VEDA also need other drivers (e.g. for netcdf) as well? The full list is somewhere in https://github.com/conda-forge/gdal-feedstock/blob/a28b550c8887cad64fd5fc43ad5a64d6c9852398/recipe/meta.yaml#L554-L567 (look around other parts of that file too).

wildintellect: That's the most important one for better vector support. We may come back later after some more experimentation with grib2 that @hrodmn has been doing but that's not common yet the way geoparquet is.

scottyhq (approved):

PR #616, opened 2025-07-30
Add support for hdf5 via libgdal-netcdf

moradology:
At the moment, GDAL doesn't actually handle `.h5` files! This appears to have been an oversight. Luckily, the fix is fairly simple


pangeo-notebook/environment.yml
```diff
@@ -43,6 +43,7 @@ dependencies:
  - gsw
  - h5netcdf
  - h5py
+ - hdf5
```
weiji14, line 46: This line can be removed, because we already depend on `netcdf4` -> `libnetcdf` -> `hdf5`.

```suggestion
```

pangeo-notebook/environment.yml
```diff
@@ -65,6 +66,7 @@ dependencies:
  - jupyter-resource-usage
  - jupyter-sshd-proxy
  - kerchunk
+ - libgdal-hdf5
```
weiji14, line 69: _Originally posted by @wildintellect in [#36](https://github.com/NASA-IMPACT/pangeo-notebook-veda-image/issues/36#issuecomment-3137683085)_
> 
> Open question what about netcdf support in gdal? 

@wildintellect, do you want `libgdal-netcdf` also? Since [`libgdal-netcdf` depends on `libgdal-hdf5` and `libgdal-hdf4`](https://github.com/conda-forge/gdal-feedstock/blob/891251423f09b5ae41c7acedebc82fd386648e75/recipe/meta.yaml#L464-L467), we could switch this line to:

```suggestion
 - libgdal-netcdf
```

weiji14: Ok, I've applied the change

weiji14:
New tag with this change available at [`2025.08.14`](https://github.com/pangeo-data/pangeo-docker-images/releases/tag/2025.08.14)

PR #642, opened 2026-03-01
Bump python from 3.13 to 3.14

weiji14:
Bumps [python](https://github.com/python/cpython) from 3.12.12 to 3.14.3.
  - [Release notes](https://github.com/python/cpython/releases/tag/v3.12.12)
  - [Changelog](https://docs.python.org/3.14/whatsnew/3.14.html)
  - [Commits](https://github.com/python/cpython/compare/v3.12.12...v3.14.3)

Previous PR at https://github.com/pangeo-data/pangeo-docker-images/pull/643

PR #643, opened 2026-03-06
Bump python from 3.12 to 3.13 for base-, pangeo- and pytorch-notebook

weiji14:
Bumps [python](https://github.com/python/cpython) from 3.12.12 to 3.13.12 for `base-notebook`, `pangeo-notebook` and `pytorch-notebook` only (`ml-notebook` stays on Python 3.12).
  - [Release notes](https://github.com/python/cpython/releases/tag/v3.13.12)
  - [Changelog](https://docs.python.org/3.13/whatsnew/3.13.html)
  - [Commits](https://github.com/python/cpython/compare/v3.12.12...v3.13.12)

Previous PR at https://github.com/pangeo-data/pangeo-docker-images/pull/556


scottyhq:
Personally I wouldn't mind just abandoning the tensorflow image going forward to only have pytorch :) 3.12 is starting to feel old 😅

weiji14:
I don't mind either since I'm not using tensorflow, but there are ~20 2i2c clusters using it from looking at https://github.com/search?q=repo%3A2i2c-org%2Finfrastructure+pangeo%2Fml-notebook+language%3AYAML&type=code&l=YAML.

Not sure if it's worth dropping it temporarily, and then picking back on it later when it's been packaged for Python 3.13?

scottyhq:
One way forward is to break from the "all images use the same python version as defined in base image" approach. I believe conflicted package versions are resolved with a left to right order, so the following changes would only use python=3.12 in the ml-image:

(base-notebook/environment.yaml -> python=3.13, ml-notebook/environment.yml -> python=3.12) -> python 3.12
```bash
cd ml-notebook
conda-lock lock -f ../base-notebook/environment.yml -f ../pangeo-notebook/environment.yml -f environment.yml -p linux-64`
```

So the steps would be 

1. add python=3.12 pin to https://github.com/pangeo-data/pangeo-docker-images/blob/ce4b6ba0882d01d92cc3f7f2dcbe83dba0963eeb/ml-notebook/environment.yml#L7-L8

2. reverse order of environment.yml input files in:

https://github.com/pangeo-data/pangeo-docker-images/blob/ce4b6ba0882d01d92cc3f7f2dcbe83dba0963eeb/Makefile#L32

and 

https://github.com/pangeo-data/pangeo-docker-images/blob/ce4b6ba0882d01d92cc3f7f2dcbe83dba0963eeb/.github/workflows/CondaLock.yml#L42

weiji14:
Thanks for the tip! Have opened #657 to reverse the environment.yml order for conda-lock on ml-notebook. Take a look at that one, and we can merge that before upgrading to Python 3.13 here.

scottyhq (approved):
woohoo! i say merge and tag, and then we can revisit https://github.com/pangeo-data/pangeo-docker-images/pull/642 ;)

weiji14:
Cool, I'll go for lunch and then push the tag later after the builds have completed on main/master.

PR #649, opened 2026-04-01
Secure some GH Actions and use ubuntu-slim for some workflows

weiji14:
Gradually securing our GitHub Actions workflows a bit, tackling some low-hanging fruit for now identified by [Zizmor](https://docs.zizmor.sh/audits):

- [x] https://docs.zizmor.sh/audits/#dependabot-cooldown
- [x] https://docs.zizmor.sh/audits/#template-injection
- [x] https://docs.zizmor.sh/audits/#unpinned-uses
- [ ] there are more, but will tackle them in the future as they are more involved

Also switching from `ubuntu-latest` to `ubuntu-slim` for some lightweight workflows (https://github.blog/changelog/2025-10-28-1-vcpu-linux-runner-now-available-in-github-actions-in-public-preview/) to save a bit of energy.

Note: I've also changed some settings at https://github.com/pangeo-data/pangeo-docker-images/settings/actions to tighthen up the permissions, so it'll be harder for breaches in general.

### DIFF
--- a/.github/workflows/WatchCondaForge.yml
+++ b/.github/workflows/WatchCondaForge.yml
@@ -22,7 +22,7 @@ jobs:
 
       - name: Get Latest pangeo-notebook Metapackage Version
         id: latest_version
-        uses: jacobtomlinson/gha-anaconda-package-version@0.1.4
+        uses: jacobtomlinson/gha-anaconda-package-version@f5d2c85fa7353b97ce642c233499724caa82328d # 0.1.4
         with:
           org: "conda-forge"
           package: "pangeo-notebook"
@@ -30,7 +30,7 @@ jobs:
 
       - name: Find and Replace pangeo-notebook
         id: find_replace
-        uses: jacobtomlinson/gha-find-replace@3.0.5
+        uses: jacobtomlinson/gha-find-replace@f1069b438f125e5395d84d1c6fd3b559a7880cb5 # 3.0.5
         with:
           include: "base-notebook/environment.yml"
           find: "pangeo-notebook=.*"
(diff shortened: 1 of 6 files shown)


scottyhq:
Cool, had not heard of zizmor before! Agreed this seems like a good idea, especially for repositories like this that build on PRs and do things like slash commands :) 

> Also switching from ubuntu-latest to ubuntu-slim for some lightweight workflows (https://github.blog/changelog/2025-10-28-1-vcpu-linux-runner-now-available-in-github-actions-in-public-preview/) to save a bit of energy. 

great!

.github/workflows/WatchCondaForge.yml
```diff
@@ -22,15 +22,15 @@ jobs:
 
       - name: Get Latest pangeo-notebook Metapackage Version
         id: latest_version
-        uses: jacobtomlinson/gha-anaconda-package-version@0.1.4
+        uses: jacobtomlinson/gha-anaconda-package-version@f5d2c85fa7353b97ce642c233499724caa82328d # 0.1.4
```
scottyhq, line 25: I didn't realize at first that the inline semver (`#0.1.4`) comment is important:

> Dependabot updates the version documentation of GitHub Actions when the comment is on the same line, such as `actions/checkout@<commit> #<tag or link>` or `actions/checkout@<tag> #<tag or link>`.

(from https://docs.github.com/en/actions/reference/security/secure-use#keeping-the-actions-in-your-workflows-secure-and-up-to-date)

scottyhq (approved):

.github/workflows/WatchCondaForge.yml
weiji14: Yeah, it's pretty neat that dependabot updates the tag too!

PR #650, opened 2026-04-28
Use icechunk 2 and latest zarr-python

jsignell:
Icechunk v2 is required for v2 stores, but can still be used for reading and writing v1 stores as well.


pangeo-notebook/packages.txt
```diff
@@ -516,268 +521,269 @@ networkx==3.6.1
...
-sqlite==3.51.2
+sparse==0.18.0
+spdlog==1.17.0
+spirv-tools==2026.1
+sqlalchemy==2.0.49
+sqlite==3.53.0
 stack_data==0.6.3
-stackstac==0.5.1
-starlette==0.50.0
+stackstac==0.5.0
```
jsignell: Flagging this downgrade.

pangeo-notebook/packages.txt
```diff
@@ -516,268 +521,269 @@ networkx==3.6.1
...
 threadpoolctl==3.6.0
-tifffile==2026.1.14
-tiledb==2.30.0
+tifffile==2026.4.11
+tiledb==2.30.1
 tiledb-py==0.36.0
-timezonefinder==8.2.1
-tinycss2==1.5.1
+timezonefinder==8.2.2
+tinycss2==1.4.0
```
jsignell, line 719: Another downgrade

pangeo-notebook/packages.txt
```diff
@@ -1,123 +1,128 @@
...
+bokeh==3.9.0
 boltons==25.0.0
-boto3==1.40.70
-botocore==1.40.70
+boto3==1.42.91
+botocore==1.42.91
 bottleneck==1.6.0
 bounded-pool-executor==0.0.3
-bqplot==0.12.45
+bqplot==0.12.37
```
jsignell: flagging this downgrade

pangeo-notebook/packages.txt
weiji14: Hmm, so this is because `stackstac=0.5.1` pins to `pandas<3` (https://github.com/gjoseph92/stackstac/pull/254). So we'll need to decide whether to pick `stackstac=0.5.1` or `pandas=3`.

pangeo-notebook/packages.txt
weiji14: These ones should be ok to ignore.

pangeo-notebook/packages.txt
weiji14: Downgrade happening because `bqplot>=0.12.38,<=0.12.46` is pinning to `pandas<3`. I was gonna recommend the new release `bqplot=0.12.46` that just came out, but it appears the backport for pandas 3 support wasn't applied correctly (see https://github.com/bqplot/bqplot/pull/1700#issuecomment-4335269908) 😅.

weiji14 (commented):
Thanks @jsignell. Looks like the downgrades you mentioned are due to packages pinning to `pandas<3`. Maybe stay on `pandas=2` for now?

pangeo-notebook/packages.txt
jsignell: I think the image could reasonably drop stackstac since it is not being maintained https://github.com/gjoseph92/stackstac/issues/270#issuecomment-4122936829

jsignell:
> Thanks @jsignell. Looks like the downgrades you mentioned are due to packages pinning to `pandas<3`. Maybe stay on `pandas=2` for now?

Yeah let me put that pin and see what the diff looks like. Might also be reasonable to wait for bqplot to implement pandas 3 support if they are working on it.

pangeo-notebook/packages.txt
weiji14: Let's hold off on dropping it until we really do need pandas=3. I'm a bit reluctant to drop `stackstac` being a heavy user of it in the past (`odc-stac` works, but it's always felt more cumbersome). Also hoping that https://github.com/developmentseed/lazycogs might mature a bit in this STAC->xarray space in a few months to act as another alternative.

weiji14 (approved):
Didn' see any significant downgrades anymore, should be ok to merge once CI passes.

Issue #653, opened 2026-05-29
2026.05.29 tag published stale(?) notebook images


tylanderson:
**Describe the bug**
Release tag 2026.05.29 publishes inconsistent/stale Docker images. The source for this tag expects icechunk v2+ (and CUDA 13 for pytorch-notebook), but published images do not contain these:

- pangeo-notebook:2026.05.29 contains icechunk 1.1.17
- ml-notebook and pytorch-notebook also show missing conda env notebook at runtime (only base env appears available)
- pangeo-notebook, ml-notebook, and pytorch-notebook all appear to be stale from the created date of the image (base images appear to be new builds)

**To Reproduce**

Icechunk version:
`docker run --rm pangeo/pangeo-notebook:2026.05.29 conda list icechunk`

Conda notebook env missing:
`docker run --rm pangeo/ml-notebook:2026.05.29 conda info --envs`
`docker run --rm pangeo/pytorch-notebook:2026.05.29 conda info --envs`

Creation date:
`docker image inspect pangeo/pangeo-notebook:2026.05.29 --format '{{.Created}}'`
`docker image inspect pangeo/ml-notebook:2026.05.29 --format '{{.Created}}'`
`docker image inspect pangeo/pytorch-notebook:2026.05.29 --format '{{.Created}}'`

**Expected behavior**

**Docker Image Version: 2026.05.29**

**Infrastructure:**
- personal computer (Apple Silicon)

scottyhq:
> Conda notebook env missing: 

Thanks for pointing this out.

It looks like the build step is actually failing at the conda environment creation step but https://github.com/docker/build-push-action doesn't surface the error so the build proceeds and pushes the bad image

https://github.com/pangeo-data/pangeo-docker-images/actions/runs/26609041435/job/78411261316

> #21 33.05 conda_lock.errors.PlatformValidationError: The lockfile conda-lock.yml does not contain a solution for the current platform linux-aarch64. The lockfile only contains solutions for the following platforms: linux-64. In order to add support for linux-aarch64, you must regenerate the lockfile. Either add the following section to your environment.yml:

Really the action shouldn't try to build the ML environments on linux-aarch64 (I'm not sure either GPU environment is possible on linux-aarch64  but feel free to try it out!)... I'll plan to at least update the Action so it doesn't push an image with no environment

scottyhq:
> pangeo-notebook:2026.05.29 contains icechunk 1.1.17 

[...]

PR #654, opened 2026-05-29
Don't build linux-aarch64 (linux/arm64) ML images

scottyhq:
- Partially addresses #653 to simply skip linux/arm64 builds for pytorch and ml environments
- Alternative if the conda packages all now work with linux-aarch64 somebody can follow up to add that! 
- Also updates the conda-lock environment since it's been a while

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -72,7 +72,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


weiji14 (approved):
Thanks @scottyhq. I'm afk today, so feel free to tag later. Linux-aarch64 should work with the ML packages, but I havent time to look into it this weekend.

